PR #249, opened 2014-10-22
Material Design for Bootstrap

lipis:
The project is unfinished but on a very good track: http://fezvrasta.github.io/bootstrap-material-design/

This issue will serve as an example for adding new themes.


lipis:
Outdated and not finished.. might add it in the future as a more fresh example

Issue #253, opened 2014-10-28
Include all 3rd party OAuth from gae-init-auth here


lipis:
Since the whole thing is broken down now in separate files and even in the admin console there is separate link for OAuth maybe it makes sense to include all of them here and not maintaining both files as it's not actually adding any overhead AFAIK...

What do you think?!

#### Da list!

- [x] Google
- [x] Facebook
- [x] Twitter
- [x] Bitbucket
- [x] Dropbox
- [x] GitHub
- [x] Instagram
- [x] Linkedin
- [x] Microsoft
- [x] Reddit
- [ ] StackOverflow
- [x] VK
- [x] Yahoo!

Everything should be sorted by name except for the `oauth.html` template where the providers **Google**, **Facebook** and **Twitter** always should be first and the rest in alphabetical order.

gmist:
@dhstack +1
We should migrate to OAuth 2.0. Deadline April 20, 2015.

joenorton:
+1

lipis:
Might essentially rename the https://github.com/gae-init/gae-init/blob/master/main/auth/google.py to `gae.py` and just create a usual OAuth for `google.py` as the rest.. we still have plenty of time.. 

We might need to keep the gae as with normal Google OAuth there will be now way to (I believe) to figure out if you are an admin out of the box..

gmist:
What about VK OAuth #266 ?

lipis:
yes.. VK should be merged.. not sure why I stopped.. maybe something with the https like on reddit?

lipis:
I think I'll close this one, as the two remaining ones are not that popular and they could always be found here if needed: https://github.com/lipis/gae-init-auth/blob/master/main/auth.py#L734-L879 (Someone could argue that Bitbucket and others are also not popular... but now it's too late :dancer: )

gmist:
What's the problem with Reddit OAuth?

lipis:
There is no problem with it.. I just think that it's not that popular.. :) But if you think that it's useful we can totally add it :)

gmist:
It would be nice.

lipis:
:+1: Will add it later today.. unless you want to do it earlier :D

lipis:
SO will be skipped.. for now at least :)

PR #266, opened 2014-11-18
Added VK OAuth

lipis:
Related #253


lipis:
Messed up with _git stuff_ but now I know :)

PR #276, opened 2015-01-02
Fix IS_WINDOWS constant population in run.py

sadovnychyi:
With "is" it will always return False.
http://stackoverflow.com/questions/2988017/string-comparison-in-python-is-vs

### DIFF
--- a/run.py
+++ b/run.py
@@ -67,7 +67,7 @@ ARGS = PARSER.parse_args()
 ###############################################################################
 BAD_ENDINGS = ['pyc', 'pyo', '~']
 GAE_PATH = ''
-IS_WINDOWS = platform.system() is 'Windows'
+IS_WINDOWS = platform.system() == 'Windows'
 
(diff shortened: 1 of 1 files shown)


lipis:
While with `is` not always going to return `False` (and I'm sure in this will never be an issue for that particular string) it looks like we should never rely on the fact that Python stores only one version of the string..

Bunch of interesting answers could also be found here:
http://stackoverflow.com/questions/1392433/python-why-is-hello-is-hello
http://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce

gmist:
I think that in this case it will never return ```False``` because ```platform.uname()```returns a tuple of strings constants ('Windows', 'Darwin', etc) as a global variable ```_uname_cache```, but better to be safe and use ```==``` operator. @sadovnychyi thank you!

PR #279, opened 2015-01-07
Form fields marked as optional if field not required

dhstack:
Thought this would be useful feedback for users before they got the validation message

Example:
image


lipis:
I was trying to avoid that for quite some time.. because I think it's a bit ugly.. :/ But I can see it could be useful.. not sure what to do :dancer:

dhstack:
Perhaps use a Font Awesome icon?

https://fortawesome.github.io/Font-Awesome/icons/


*--*
*David Stack*
http://www.linkedin.com/in/davidstack

dhstack:
Like this or something:
https://fortawesome.github.io/Font-Awesome/icon/exclamation-circle/

dhstack:
Added some icons so it would look like this:
image

@lipis What do you think?

lipis:
I'm guessing you are using `<sup>` there?

A better idea if using FA is to do something like this:

```html
<sup><span class="fa fa-asterisk small text-muted" title="Required"></span></sup>
```

But is still too noisy I believe..

```
<sup class="small text-muted" title="Required">*</sup>
```

or `&deg;` instead of `*`

dhstack:
@lipis you guessed correctly. I like the look of the fa asterisk better than the * since otherwise it's almost too small to even see, but muting it was a good idea.

image

dhstack:
In light of an article I just came across about form UX design, it might be better to just leave it as is (no asterisk) or mark fields as optional.

http://uxmovement.com/forms/why-users-fill-out-less-if-you-mark-required-fields/

lipis:
Yes.. marking optional fields is better idea.. and I wanted to suggest that but there is no icon for it.. I guess just `small text-muted` is enough though.. :)

lipis:
Great article @dhstack

dhstack:
@lipis I fixed the feedback form label - are there any other forms that I'm missing? The sign up is okay

lipis:
Will test/evaluate tomorrow.. thanks!

lipis:
Not too bad..
screen shot 2015-01-08 at 12 11 29

gmist:
+1

lipis:
@dhstack One more place.. https://github.com/gae-init/gae-init/blob/master/main/templates/user/user_email_field.html#L3

lipis:
Also for the recaptcha, the visible password, 

In general search for `{{field.label..` inside the `forms.html`

topless:
legit

lipis:
For anyone that is not aware of this little page: http://localhost:8080/admin/test/form/ (when running locally)

dhstack:
@lipis should be good to go now

dhstack:
@lipis should be good to go now

main/templates/user/user_email_field.html
```diff
@@ -1,6 +1,7 @@
+# import 'macro/forms.html' as forms
```
lipis: One empty line after import..

main/templates/macro/forms.html
```diff
@@ -12,9 +12,16 @@
...
+  # if not field.flags.required
+    <small class="text-muted">(optional)</small>
+  # endif
+# endmacro
+
+
 # macro input_field(field, prefix='', suffix='')
   <div class="form-group {{'has-error' if field.errors}}">
-    {{field.label(class='control-label')}}
+    {{field.label(class='control-label')}} {{field_optional(field)}}
```
lipis: Maybe putting all those in a separate line..

dhstack:
@lipis okay, should be good now

lipis:
Sorry for the delays.. but like I mentioned before it's quite a change :)

For example it looks like the reCAPTCHA might not be needing it after all.. as who would ever put recaptcha and not requiring it.. plus in our case there is no required flag for it because it's being removed if not able to use it..

http://dl.dropbox.com/u/15234/_/Screen%20Shot%202015-01-08%20at%2023.37.49.png

so lets ditch the label on reCAPCTHA

dhstack:
No worries on the delays - better to do it right the first time than have bugs later on.

Good point about the reCAPTCHA, though it is funny to see "Are you human?" labeled as optional, as it defeats the purpose :)

lipis:
it looks good.. I will test it a bit more and sleep over it.. but either case you will have to squash your commits into one (to avoid all the noise in the history).. if you're not very familiar with it.. ask away and we'll come back to it tomorrow.. gotta go

dhstack:
How do I squash all the commits into one?

lipis:
might be tricky since you are updated your master branch instead of a new one.. http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

Make sure you save your changes somewhere else in case of total failure :) worst case new branch :) gl

dhstack:
Okay, definitely messed this up. I created a [new branch](https://github.com/dhstack/gae-init/tree/optional-field-label) - should I make a new pull request or is there a way to change the branch on this one?

lipis:
I thing the main issue is that because you made these changes on your master branch and it's not in sync with the parent..

Not sure how you can rebase/sync to be one to one with the gae-init:master and your master.. Ideally I would like one commit per PR.. but in special cases we could do some exceptions.. make a new PR with one or two commits and I'll merge it..

In the future though.. try to sync the two master branches :) @topless also managed to mess up with his PRs :)

dhstack:
As requested, one commit :) I ended up creating a new branch from when I originally forked it, and that seemed to work. Since these were the only changes I made it wasn't a big deal.

Sorry for the duplicate pull requests, but at least now I know how to do one correctly. In the future I'll be sure the branches are in sync - thanks for your patience.

lipis:
Thanks for that.. You might want to sync your master now with the original master :)

PR #280, opened 2015-01-09
Styles and less compilation is handled by gulp.

topless:


run.py
```diff
@@ -540,10 +498,10 @@ def run_clean_all():
   print_out('CLEAN ALL')
   remove_file_dir([
       DIR_BOWER_COMPONENTS, DIR_NODE_MODULES, DIR_EXT, DIR_MIN, DIR_DST
-    ])
+  ])
   remove_file_dir([
       FILE_PIP_GUARD, FILE_NPM_GUARD, FILE_BOWER_GUARD
-    ])
+  ])
```
lipis: keep those where they were :)

run.py
```diff
@@ -565,7 +520,7 @@ def run_minify():
     coffees = separator.join([
         os.path.join(DIR_STATIC, script)
         for script in scripts if script.endswith('.coffee')
-      ])
+    ])
```
lipis: same hre :)

PR #281, opened 2015-01-10
Added form fields to be marked as optional if field is not required

dhstack:
Thought this would be helpful for users to know which fields are optional.

Note: This is a new version of pull request #279


lipis:
Thanks

PR #284, opened 2015-01-10
Gulp as build system for front end.

topless:


gulpfile.coffee
```diff
@@ -1,69 +1,116 @@
...
+    "#{bower_dir}/bootstrap/js/tooltip.js"
+  ]
   clean: [
-      "#{static_dir}/dst"
-      "#{static_dir}/ext"
-      "#{static_dir}/min"
-    ]
+    "#{static_dir}/dst"
+    "#{static_dir}/min"
+  ]
```
lipis: sorry but you know the convention already :dancer:

gulpfile.coffee
```diff
@@ -1,69 +1,116 @@
...
-
-run = (option) ->
-  proc = exec "python -u run.py -#{option}"
-  proc.stderr.on 'data', (data) -> process.stderr.write data
-  proc.stdout.on 'data', (data) -> process.stdout.write data
+    "#{static_dir}/dst/style/**/*.css"
+    "#{static_dir}/dst/script/**/*.js"
+    "#{root_dir}/**/*.html"
+    "#{root_dir}/**/*.py"
+  ]
```
lipis: Same here :)

main/config.py
```diff
@@ -2,7 +2,6 @@
 
 import os
 
-
```
lipis: ?

gulpfile.coffee
```diff
@@ -1,69 +1,116 @@
...
+  gulp.src(paths.scripts)
+    .pipe $.concat("script.min.js")
+    .pipe $.uglify()
+    .pipe gulp.dest("#{static_dir}/min/script/")
+
+  gulp.src(paths.bower_components)
+    .pipe $.plumber()
+    .pipe $.concat("libs.min.js")
+    .pipe $.uglify()
+    .pipe gulp.dest("#{static_dir}/min/script/")
```
lipis: And this part is where I don't like.. :( but `script`, `libs` are very much hardcoded here and it will be hard to split things in different groups

I agree that most likely in practice you don't need more than two or it could be all one.. but nevertheless I'm guessing some projects require more..

main/static/src/style/style.less
```diff
@@ -1,7 +1,9 @@
-@import "../../ext/font-awesome/less/font-awesome";
-@import "../../ext/bootstrap/less/bootstrap";
-@import "../../ext/bootstrap-sweetalert/lib/sweet-alert.less";
-@import "../../ext/nprogress/nprogress";
+@bower: "../../bower_components";
```
lipis: not winning much with this variable... same thing happened with `@ext` back in the day..

PR #289, opened 2015-01-24
Moving controllers into control

lipis:
Kind of experimentish :) @topless will like that one.. WDYT?


gmist:
+1

lipis:
The only problem with that one is that it will break bunch of things if people changed these files.. but it should happen :0

PR #293, opened 2015-01-28
Check for SemanticVersion in the latest pip first

### LINKED ISSUE #292: SemanticVersion no longer works in pip 6.0+
```shell
  File "./run.py", line 658, in <module>
    run()
  File "./run.py", line 639, in run
    print_out_update()
  File "./run.py", line 421, in print_out_update
    SemVer = pip._vendor.distlib.version.SemanticVersion
AttributeError: 'module' object has no attribute 'version'
```

lipis:
Fixes #292

### DIFF
--- a/run.py
+++ b/run.py
@@ -418,7 +418,11 @@ def check_for_update():
 
 def print_out_update():
   import pip
-  SemVer = pip.util.version.SemanticVersion
+  try:
+    SemVer = pip.util.version.SemanticVersion
+  except AttributeError:
+    SemVer = pip._vendor.distlib.version.SemanticVersion
+
   try:
     with open(FILE_UPDATE, 'r') as update_json:
       data = json.load(update_json)


run.py
```diff
@@ -418,7 +418,11 @@ def check_for_update():
 
 def print_out_update():
   import pip
-  SemVer = pip.util.version.SemanticVersion
+  try:
+    SemVer = pip._vendor.distlib.version.SemanticVersion
+  except:
```
gmist:
``` python 
except ImportError:
```

lipis: `import pip` is not party of the `try/except` so I'm guessing it will be the `AttributeError`.. no?

gmist: +1 Of course, I was wrong

lipis: Updated and also changed the order, cause in future it will be the opposite :)

PR #295, opened 2015-01-30
Change loading text to spinning loading icon

dhstack:
Added a spinning gear when the buttons are "loading" after being clicked instead of the different text.

It makes for a consistent experience across the site.

image
It spins when live

### DIFF
--- a/main/templates/macro/forms.html
+++ b/main/templates/macro/forms.html
@@ -186,3 +186,12 @@
     </div>
   </div>
 # endmacro
+
+
+# macro data_loading_text(text='Please wait..', icon='fa fa-spin fa-spinner')
+  # if icon
+    data-loading-text="<span class='{{icon}}'></span> {{text}}"
+  # else
+    data-loading-text="{{text}}"
+  # endif
+# endmacro
--- a/main/templates/admin/admin_auth.html
+++ b/main/templates/admin/admin_auth.html
@@ -17,7 +17,7 @@
           # include 'admin/bit/twitter_oauth.html'
           # include 'admin/bit/yahoo_oauth.html'
         </div>
-        <button type="submit" class="btn btn-primary btn-lg btn-block btn-loading" data-loading-text="<span class='fa fa-fw fa-spin fa-cog'></span>">
+        <button type="submit" class="btn btn-primary btn-lg btn-block btn-loading" {{forms.data_loading_text()}}>
           Update Config
         </button>
       </form>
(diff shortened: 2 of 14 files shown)


lipis:
I would add something like that in the `forms.html`

```html
# macro data_loading_text(text='Please wait..', icon='fa fa-spin fa-spinner')
  # if icon
    data-loading-text="<span class='{{icon}}'></span> {{text}}"
  # else
    data-loading-text="{{text}}"
  # endif
# endmacro

```

And then use it like `{{forms.data_loading_text()}}`

p.s. By default I would have the generic text but if someone wants only the spinner could do the: `{{forms.data_loading_text('')}}`

dhstack:
This is a much better way to implement it. I'll work on it.

main/templates/macro/forms.html
```diff
@@ -186,3 +186,12 @@ <h4 class="panel-title">
...
 # endmacro
+
+
+# macro data_loading_text(text='Please wait..', icon='fa fa-spin fa-spinner')
+  # if icon
+    data-loading-text="<span class='{{icon}}'></span> {{text}}"
+  # else
+    data-loading-text="{{text}}"
+  # endif
+# endmacro
```
lipis: end of line

lipis:
why did you close this one?!

dhstack:
I was trying to squash my commits and messed up...

dhstack:
accidentally deleted the branch, but I made a new one with the same name thinking that would work. It didn't.

lipis:
I would recommend wait until all comments remarks are there and then I could just let you know what to type in order to achieve that easily if there are no merges, because you can do that without rebase :)

In short that's one easy way of doing it: http://stackoverflow.com/a/5201642/8418

lipis:
You could also create the same branch name locally and then force push the changes and that would not close the PR..

dhstack:
So that's the approach I tried (reset --soft), but it was telling me I needed to merge when I tried to push to server. Not sure what I did wrong. I'll get it eventually. Thanks for bearing with me until I do though :)

Since I can't reopen this PR, I'll just make a new one.

gmist:
``` shell
git rebase -i master
git push -f
```
imho it is most simple and logical way for squash commits

lipis:
or that... usually @gmist is right :dancer: 
I'm not the guru when it comes to the rebasing/squshing/etc but lately the rebase worked nicely.. :)

dhstack:
@gmist I'll try that next time, thanks.

Issue #294, opened 2015-01-30
Password hashing and unique salt


dhstack:
Instead of using a static salt value (the one stored in config.CONFIG_DB.salt - see [why salt reuse is ineffective](https://crackstation.net/hashing-security.htm#ineffective)), the user key, and user signup time for hashing users' passwords, it might be better to generate a single, randomly generated, salt value for each user using the methodology described in [this article](https://crackstation.net/hashing-security.htm#properhashing).

Then, all you'd need is a single `password_salt` property for each user in the datastore and simplify the `password_hash()` function in util.py to something like:
```
def password_hash(password, salt):
  m = hashlib.sha256()
  m.update(password.encode('utf-8'))
  m.update(salt)
  return m.hexdigest()
```

Of course, the `get_user_db_from_email()` function in auth/auth.py would also need to be slightly modified to pull out the password_salt value from the user datastore when checking the submitted password.

Also, when a user is created (or the password updated) a new salt would need to be saved to the user datastore entity. [The article](https://crackstation.net/hashing-security.htm#properhashing) suggests the salt should be at least the length of the hash, so something like this should work when creating/updating a user password:
`password_salt = os.urandom(32)` (length for sha256 from [hash.digest_size](https://docs.python.org/2/library/hashlib.html)).

dhstack:
Documentation for [os.urandom()](https://docs.python.org/2/library/os.html#miscellaneous-functions)

joernhees:
... said this when it was introduced... https://github.com/gae-init/gae-init/pull/201#issuecomment-52425144

IIRC, the rational back then was that the DB key is already added to the global salt and "sort of random". That was liked more than adding another field to the user...

lipis:
It was something among those lines sure.. but the thing is that in the end we are using a *salt* per user.. https://github.com/gae-init/gae-init/blob/master/main/util.py#L219-L220

Agree is not that random as it's essentially the timestamp.. but the milliseconds are pretty random, plus we have the key :)

Even if you know how it's produced it's impossible to guess and if you really really want to.. you can add a random number of milliseconds from this range `[-99999..99999]` to the created date, whenever you actually need to reset something, on top of just reseting the general salt which is also included regardless of the solution just to make sure that with one move you can *force* the users to reset their passwords without updating the *salts* per user..

Practically two users with the same password will have different hash and if in any way the database will be leaked or something in both scenarios the salt or creation date or the general salt will be compromised too...

[...]

PR #202, opened 2014-08-17
Used Flask-RESTful for API

gmist:
Closes #162

PR #297, opened 2015-02-02
Refactor user password salting

dhstack:
In reference to issue #294

### DIFF
--- a/main/model/user.py
+++ b/main/model/user.py
@@ -22,6 +22,7 @@ class User(model.Base):
   verified = ndb.BooleanProperty(default=False)
   token = ndb.StringProperty(default='')
   password_hash = ndb.StringProperty(default='')
+  password_salt = ndb.StringProperty(default='')
 
   def has_permission(self, perm):
     return self.admin or perm in self.permissions
--- a/main/profile.py
+++ b/main/profile.py
@@ -104,7 +104,7 @@ def profile_password():
     new_password = form.new_password.data
     if new_password or old_password:
       if user_db.password_hash:
-        if util.password_hash(user_db, old_password) != user_db.password_hash:
+        if util.password_hash(old_password, user_db.password_salt) != user_db.password_hash:
           form.old_password.errors.append('Invalid current password')
           errors = True
       if not errors and old_password and not new_password:
@@ -112,7 +112,8 @@ def profile_password():
         errors = True
 
       if not (form.errors or errors):
-        user_db.password_hash = util.password_hash(user_db, new_password)
+        user_db.password_salt = util.password_salt()
+        user_db.password_hash = util.password_hash(new_password, user_db.password_salt)
         flask.flash('Your password has been changed.', category='success')
 
     if not (form.errors or errors):
--- a/main/profile.py
+++ b/main/profile.py
@@ -104,7 +104,7 @@ def profile_password():
     new_password = form.new_password.data
     if new_password or old_password:
       if user_db.password_hash:
-        if util.password_hash(old_password, user_db.password_salt) != user_db.password_hash:
+        if util.password_hash(user_db, old_password) != user_db.password_hash:
           form.old_password.errors.append('Invalid current password')
           errors = True
       if not errors and old_password and not new_password:
@@ -113,7 +113,7 @@ def profile_password():
 
       if not (form.errors or errors):
         user_db.password_salt = util.password_salt()
-        user_db.password_hash = util.password_hash(new_password, user_db.password_salt)
+        user_db.password_hash = util.password_hash(user_db, new_password)
         flask.flash('Your password has been changed.', category='success')
 
     if not (form.errors or errors):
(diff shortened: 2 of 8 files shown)


main/util.py
```diff
@@ -214,16 +216,38 @@ def create_name_from_email(email):
...
 
 
+def password_salt(n_bytes=32):
+  return binascii.hexlify(os.urandom(n_bytes))
+
+
+# Taken from Django Source Code
+# Helps prevent timing attacks
+# See http://carlos.bueno.org/2011/10/timing.html
+def constant_time_compare(val1, val2):
```
lipis: Regardless of the fact that this PR is not going to make it (as discussed in #294) I don't see any point of doing that here as we are comparing hashes and the hashes always have the same length.. am I missing something?

dhstack: @lipis This function secures against a [Timing Attack](https://en.wikipedia.org/wiki/Timing_attack). In python, == returns false as soon as there is a single difference in the values. Thus, an attacker could theoretically [crack the password one letter at a time](http://codahale.com/a-lesson-in-timing-attacks/). This function always takes the same amount of time to finish, regardless of whether the two strings match or not, which is why Django uses it.

Even if you don't accept this whole PR, it might be worth adding this in.

lipis: yes.. you are right :)

but at the moment you can have only 8 attempts (https://github.com/gae-init/gae-init/blob/master/main/config.py#L32) before the reCAPTCHA will kick in.. so I don't see any reason to optimise that part and I think it will make more sense with native apps rather than webapps as precise timing depending on many other things especially when milliseconds count :)

Also regarding the custom login in general there is a very nice post by Jeff Atwood on how they did on Discourse https://blog.codinghorror.com/the-god-login/ and one thing that I'm considering adding in the near future is the server side time delay after consecutive wrong attempts.. to get rid of the reCAPTCHA :)

dhstack: I like the reCAPTCHA defense, but the way I see it, the more layers of security you have, the better off you are.

What if a user of gae-init decides to turn off reCAPTCHA? They may unknowingly open themselves up to the timing attack.

While I would agree it may be more likely in native apps, it is theoretically possible:
>In 2003, Boneh and Brumley demonstrated a practical network-based timing attack on SSL-enabled web servers, based on a different vulnerability having to do with the use of RSA with Chinese remainder theorem optimizations. The actual network distance was small in their experiments, but the attack successfully recovered a server private key in a matter of hours.

That's a great article. I particularly like the section on helping users choose better passwords, where they don't allow the 10,000 most common. That'd be cool to implement.

You'd think there'd be a standard log in library by now with all the best practices ready to go for anyone creating a website...regardless, I think gae-init is off to a great start for becoming a solid standard.

dhstack: The caps locks notification from the article was also a nice touch

dhstack: @lipis Also, another reason to add the constant time comparison: I haven't configured reCAPTCHA yet and can attempt any number of login attempts. Thus it is not as secure right out of the box.

lipis: Actually on a second thought.. yes you are right if you trying to crack the actual password if we were storing the plain text password.. but but but we are not comparing the naked passwords.. and by adding or removing random characters we are not exposing anything.. the times will be completely random and in practice the comparison will be ending very very fast, after comparing maximum the first couple of characters.. :) and you will achieve nothing.. 

(you can try it yourself by comparing only the the sha1 values by adding the one character at the time to get to the actual password..)

dhstack: Good point. Even if an attacker figures out the hash of the password, they'd still need the salt to make use of it.

lipis:
Also to comment about the reCAPTCHA if you don't have it enabled.. think twice about enabling the custom login :) (At least before the server time delay might sooner or later arrive automagically)

dhstack:
I didn't even know that not having the reCAPTCHA enabled was a potential security issue (thus why I wanted to change the password checking function). The site is still in development, but thanks for the heads up.

I'll be sure to enable the reCAPTCHA before it goes live.

Any idea when the server time delay feature might be ready?

lipis:
Like I mentioned before I'll pass on this one this time as more important things are coming #202 soon :)

main/profile.py
```diff
@@ -112,6 +112,7 @@ def profile_password():
         errors = True
 
       if not (form.errors or errors):
+        user_db.password_salt = util.password_salt()
```
lipis: you could simply use `util.uuid()` whenever you need a salt or anything random... :)

lipis: and there is no need for extra functions

main/model/user.py
```diff
@@ -22,6 +22,7 @@ class User(model.Base):
   verified = ndb.BooleanProperty(default=False)
   token = ndb.StringProperty(default='')
   password_hash = ndb.StringProperty(default='')
+  password_salt = ndb.StringProperty(default='')
```
lipis: `default=util.uuid()`

main/profile.py
dhstack: That's a good idea, thanks :)

PR #300, opened 2015-02-06
Refactoring fields

lipis:
I cam across this issue when I wanted to insert, for example, a new datetime field into the User model and couldn't use the `api.fields.DateTimeField` as the `fields` was already imported from `restful`. Not sure if that's the best solution.. WDYT? @gmist

### DIFF
--- a/main/api/fields.py
+++ b/main/api/fields.py
@@ -3,6 +3,7 @@
 import urllib
 
 from flask.ext.restful import fields
+from flask.ext.restful.fields import *
 
 
 class BlobKey(fields.Raw):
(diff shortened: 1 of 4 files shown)


gmist:
[Something similar](https://github.com/gmist/my-gae-init/pull/4) was proposed by @topless

lipis:
Right.. I had a little dejavu I have to admit.. :) so not really sure how to proceed..

gmist:
I think it looks not bad

lipis:
will proceed now like this as I need it to make it work.. if there is anything better going to happen in the future let's see..

PR #301, opened 2015-02-07
Ability to filter by a (urlsafe) ndb key with different button text

dhstack:
For anyone who wants to dynamically generate the filter buttons on the fly using a KeyProperty from a linked model. This is how I'm using it:

```
<div class="btn-group btn-group-sm">
    <button type="button" class="btn btn-info" disabled >Genre</button>
    # for genre_db in genre_dbs
      {{utils.filter_by_link('genre_key', genre_db.key.urlsafe(), button_text=genre_db.key.get().name)}}
    # endfor
</div>
```

Which renders:
image

### DIFF
--- a/main/templates/macro/utils.html
+++ b/main/templates/macro/utils.html
@@ -11,12 +11,14 @@
 # endmacro
 
 
-# macro filter_by_link(property, value, icon=None, ignore='cursor', is_list=False, hash=None)
+# macro filter_by_link(property, value, icon=None, ignore='cursor', is_list=False, hash=None, label=None)
   # set value = '%s' % value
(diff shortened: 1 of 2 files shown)


main/templates/macro/utils.html
```diff
@@ -11,12 +11,14 @@
 # endmacro
 
 
-# macro filter_by_link(property, value, icon=None, ignore='cursor', is_list=False, hash=None)
+# macro filter_by_link(property, value, icon=None, ignore='cursor', is_list=False, hash=None, button_text=None)
```
lipis: I would just call it `label`

lipis:
Would you squash them?

PR #302, opened 2015-02-07
Refactoring *_FIELDS

lipis:
Is there any particular reason on why we shouldn't do something like this?


lipis:
@gmist wdyt?

gmist:
I see no any serious reasons to against it, +1

lipis:
Cool!

PR #303, opened 2015-02-07
Using the verbose names where possible

lipis:
Experimental..

Might look confusing or useless.. but I like the idea to make use of the `verbose_name`


gmist:
+1

main/control/admin.py
```diff
@@ -29,20 +29,20 @@ def admin():
...
-  check_unique_email = wtforms.BooleanField('Check for uniqueness of the verified emails')
-  email_authentication = wtforms.BooleanField('Email authentication for sign in/sign up')
-  feedback_email = wtforms.StringField('Feedback Email', [wtforms.validators.optional(), wtforms.validators.email()], filters=[util.email_filter])
-  flask_secret_key = wtforms.StringField('Flask Secret Key', [wtforms.validators.optional()], filters=[util.strip_filter])
-  notify_on_new_user = wtforms.BooleanField('Send an email notification when a user signs up')
-  recaptcha_private_key = wtforms.StringField('Private Key', filters=[util.strip_filter])
-  recaptcha_public_key = wtforms.StringField('Public Key', filters=[util.strip_filter])
-  salt = wtforms.StringField('Salt', [wtforms.validators.optional()], filters=[util.strip_filter])
-  verify_email = wtforms.BooleanField('Verify user emails')
+  analytics_id = wtforms.StringField(model.Config.analytics_id._verbose_name, filters=[util.strip_filter])
```
gmist: maybe something like this?
``` python
import model.Config as cfg
...
 analytics_id = wtforms.StringField(cfg.analytics_id._verbose_name, filters=[util.strip_filter])
...
```

lipis: It goes against my religion :) but if it will happen it will happen I guess only in that file.. as its not tha messy elsewhere..

PR #304, opened 2015-02-07
Refactoring API

lipis:
To be more consistent and it will scale better when deeper than one level structure for the API


main/control/user.py
```diff
@@ -21,7 +21,7 @@
 ###############################################################################
 # User List
 ###############################################################################
-@app.route('/admin/user/')
+@app.route('/admin/users/')
```
lipis: Also to be consistent with the REST

lipis:
I think I'll go for that one.. any strong opinions on for or against it?

nazroll:
+1
this adds context.

PR #308, opened 2015-02-11
Moved error handling to control module

lipis:


main/api/helpers.py
```diff
@@ -15,20 +15,24 @@ def unauthorized(self, response):
...
-      e.code = 500
-      e.name = e.description = 'Internal Server Error'
-    return util.jsonpify({
-        'status': 'error',
-        'error_code': e.code,
-        'error_name': util.slugify(e.name),
-        'error_message': e.name,
-        'error_class': e.__class__.__name__,
-        'description': e.description,
-      }), e.code
```
lipis: I didn't managed to reuse that part when removed the error handling from the `main.py` due to some circular import I would guess.. if there is any way to avoid it could restore that file @gmist

PR #309, opened 2015-02-12
Overriding existing fields

lipis:
Is there anything that could terribly wrong here..? Just to keep it less confusing with `Integer` and `IntegerField`


lipis:
@gmist WDYT?

gmist:
+1 It will help to avoid confusion (like choice between Integer/IntegerField or DateTime/DateTimeField).

lipis:
That was my point exactly.. thanks

PR #311, opened 2015-02-13
Google OAuth2

gmist:
You can try it [here](https://my-gae-init.appspot.com/signin/).

### DIFF
--- a/main/auth/gae.py
+++ b/main/auth/gae.py
@@ -1,0 +1,49 @@
+# coding: utf-8
+
+from __future__ import absolute_import
+
+from google.appengine.api import users
+import flask
+
+import auth
+import model
+import util
+
+from main import app
+
+
+@app.route('/signin/gae/')
+def signin_gae():
+  auth.save_request_params()
+  gae_url = users.create_login_url(flask.url_for('gae_authorized'))
+  return flask.redirect(gae_url)
+
+
+@app.route('/_s/callback/gae/authorized/')
+def gae_authorized():
+  gae_user = users.get_current_user()
+  if gae_user is None:
+    flask.flash('You denied the request to sign in.')
+    return flask.redirect(util.get_next_url())
+
+  user_db = retrieve_user_from_gae(gae_user)
+  return auth.signin_user_db(user_db)
+
+
+def retrieve_user_from_gae(gae_user):
+  auth_id = 'federated_%s' % gae_user.user_id()
+  user_db = model.User.get_by('auth_ids', auth_id)
+  if user_db:
+    if not user_db.admin and users.is_current_user_admin():
+      user_db.admin = True
+      user_db.put()
+    return user_db
+
+  return auth.create_user_db(
+      auth_id=auth_id,
+      name=util.create_name_from_email(gae_user.email()),
+      username=gae_user.email(),
+      email=gae_user.email(),
+      verified=True,
+      admin=users.is_current_user_admin(),
+    )
--- a/main/templates/admin/bit/google_oauth.html
+++ b/main/templates/admin/bit/google_oauth.html
@@ -1,0 +1,11 @@
+{{
+  forms.panel_fields(
+      'Google',
+      (form.google_client_id, form.google_client_secret),
+      '''
+        Redirect URI for <a href="https://console.developers.google.com/project/%s/apiui/credential" target="_blank">Google application</a>:
+        <em>%s</em><br>
+        Do not forget to enable "Google+ API" in your <a href="https://console.developers.google.com/project/%s/apiui/api" target="_blank">console</a>.
+      ''' % (config.APPLICATION_ID, url_for('google_authorized', _external=True), config.APPLICATION_ID)
+    )
+}}
--- a/main/templates/auth/auth.html
+++ b/main/templates/auth/auth.html
@@ -24,12 +24,13 @@
 
         # set is_icon = config.CONFIG_DB.has_email_authentication
         <div class="text-center">
-          {{utils.signin_button('Google', 'btn-google', 'fa-google-plus', google_signin_url, is_icon)}}
+          {{utils.signin_button('GAE' if config.CONFIG_DB.has_google else 'Google', 'btn-google', 'fa-google', gae_signin_url, is_icon)}}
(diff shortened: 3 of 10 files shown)


nazroll:
this would help with context to differentiate between the GAE Users service and Google+ Oauth2 implementation. I was initially confused between the two.

+1

joenorton:
This looks great. Just implemented it but ran into a speed bump, and I wanted to reach out to see if I'm doing something silly. I can't quite get it working when testing on localhost.

The error says 'Error: redirect_uri_mismatch' -- it is due to the redirect uri on the Google callback. When setting up the client ID you can input the redirect URI  and I have tried variations of locahost and 127.0.0.1 but I have yet to get it to work when testing locally.

When I googled it, I found this guy (http://stackoverflow.com/questions/21888461/flask-oauthlib-server-doesnt-play-well-with-requests-oauthlib-client) saying that localhost maybe is not excepting the query paramets due to it being a POST request, as opposed to GET (which is how it looks like the oauth is callback URL is done).

Have you guys run into this when running locally? Does something have to be done in oauthlib? or is there a 'magic redirect URI' that does the trick (in in the client ID settings)?

Thanks guys,

Joe

main/templates/admin/bit/google_oauth.html
```diff
@@ -0,0 +1,11 @@
+{{
+  forms.panel_fields(
+      'Google',
+      (form.google_client_id, form.google_client_secret),
+      '''
+        Callback URL for <a href="https://console.developers.google.com/project/%s/apiui/credential">Google application</a>:
```
lipis: It says `Redirect URI`.. I'm usually using exactly what is says there to find it easier:

lipis: Also `target="_blank"`

main/templates/admin/bit/google_oauth.html
```diff
@@ -0,0 +1,11 @@
...
+        <em>%s</em><br>
+        Do not forget to enable "Google+ API" in your <a href="https://console.developers.google.com/project/%s/apiui/api">console</a>.
```
lipis: `target="_blank"`

main/templates/auth/auth.html
```diff
@@ -24,12 +24,13 @@
 
         # set is_icon = config.CONFIG_DB.has_email_authentication
         <div class="text-center">
-          {{utils.signin_button('Google', 'btn-google', 'fa-google-plus', google_signin_url, is_icon)}}
+          {{utils.signin_button('GAE', 'btn-google', 'fa-google', gae_signin_url, is_icon)}}
```
lipis: Even though internally it is GAE I don't really like it GAE here, because noone will know that except the devs.. maybe just `Google` and the other one `Google+` for the labels?

lipis:
@joenorton Worked perfectly fine for me.. :) Did you figure out what was wrong?

main/templates/auth/auth.html
lipis: On a second thought.. maybe: `'GAE' if config.CONFIG_DB.has_google else 'Google'`

joenorton:
Oddly enough, I did not get it working on localhost.
I tried that URL provided, and get a bizarre error now:
'''
TypeError: __init__() got an unexpected keyword argument 'context'

File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 1166, in do_open

h = http_class(host, timeout=req.timeout, **http_conn_args)
TypeError: __init__() got an unexpected keyword argument 'context'
'''

Could it be because I am using python 2.7.9?

Whats weird is I went ahead and pushed it to production and it works fine there.

lipis:
Very strange.. I have the same version, so it shouldn't be Python.. and it works perfectly locally.

gmist:
@joenorton What version of the GAE are you using?

main/templates/macro/utils.html
```diff
@@ -70,8 +70,10 @@
     <span class="fa fa-fw fa-facebook" title="Facebook"></span>
   # elif auth_id.startswith('github')
     <span class="fa fa-fw fa-github" title="GitHub"></span>
-  # elif auth_id.startswith('federated')
+  # elif auth_id.startswith('google')
     <span class="fa fa-fw fa-google-plus" title="Google"></span>
+  # elif auth_id.startswith('federated')
+    <span class="fa fa-fw fa-google" title="GAE"></span>
```
lipis: We have space here.. so we could do `Google App Engine`

main/templates/admin/bit/google_oauth.html
```diff
@@ -0,0 +1,11 @@
+{{
+  forms.panel_fields(
+      'Google',
+      (form.google_client_id, form.google_client_secret),
+      '''
+        Redirect URI for <a href="https://console.developers.google.com/project/%s/apiui/credential" target="_blank">Google application</a>:
+        <em>%s</em><br>
+        Do not forget to enable "Google+ API" in your <a href="https://console.developers.google.com/project/%s/apiui/api">console</a>.
```
lipis: `target="_blank"`

main/auth/gae.py
```diff
@@ -0,0 +1,49 @@
+# coding: utf-8
+
+from __future__ import absolute_import
```
lipis: I think that one was needed because the filename was called `google.py` and were some conflicts with it.. if that's not the case we could get rid of it :)

(Note to self: read about it)

gmist: In this case it is necessary for correct import
``` python
from google.appengine.api import users
```
Because we import ```gae.py``` (in ```__init__.py```) through relative import
``` python
from .gae import *
```

joenorton:
Alright, finally figured it out and got it working. I was using GAE 1.9.17, now I'm on 1.9.18 and it's working.

Looks like it was a bug in fancy_urllib (when using python 2.7.9), part of the GAE sdk.
https://code.google.com/p/googleappengine/issues/detail?id=11536

I input the patch, but for some reason couldn't ever get it to recognize it. 

Ended up re-installing the whole google-cloud-sdk, GAE upgraded to 1.9.18, and now it's working on localhost!

So mystery solved. Working perfect!

main/auth/gae.py
lipis: thanks!

lipis:
:100:

PR #312, opened 2015-02-14
Added clickable rows support

lipis:


lipis:
One problem that I can see is that this should not used for public pages as it's not that obvious that they are links (for the SEO). But on the other hand.. if something is going to be public or indexable, `<table>` is not recommended anyway :D

main/templates/user/user_list.html
```diff
@@ -97,13 +97,13 @@
...
-          <tr id="{{user_db.key.urlsafe()}}" class="{{'text-muted' if not user_db.active}}">
+          <tr id="{{user_db.key.urlsafe()}}" class="row-link {{'text-muted' if not user_db.active}}" data-href="{{url_for('user_update', user_id=user_db.key.id())}}">
             <td class="name">
-              <label>
+              <label class="not-link">
                 <input type="checkbox" name="user_db" value="{{user_db.key.urlsafe()}}">
                 <img src="{{user_db.avatar_url_size(80)}}" alt="Avatar of {{user_db.name}}">
               </label>
-              <a href="{{url_for('user_update', user_id=user_db.key.id())}}">{{user_db.name}}</a>
+              <span>{{user_db.name}}</span>
```
dhstack: @lipis why does this need the span tags?

lipis: Because if you don't it looks like this
screen shot 2015-03-21 at 22 01 32

Instead of
screen shot 2015-03-21 at 22 01 20

dhstack: So it does. Missed that minor detail when I compared the two.

PR #321, opened 2015-03-01
Use Gulp for building frontend + added source maps for scripts and styles

gmist:
screen shot 2015-03-01 at 19 56 38

Based on @topless ideas (#284). Usage:

``` shell
git clone https://github.com/gmist/my-gae-init.git
cd my-gae-init
git checkout -b source_maps origin/source_maps
npm install
gulp build
gulp
```

P.S. Needs testing on Windows

### DIFF
--- a/gulpfile.coffee
+++ b/gulpfile.coffee
@@ -1,69 +1,8 @@
-gulp = require('gulp')
-$ = require('gulp-load-plugins')()
-main_bower_files = require 'main-bower-files'
-del = require 'del'
-exec = require('child_process').exec
-minimist = require 'minimist'
+gulp = require('gulp-help') require 'gulp'
+requireDir = require('require-dir') './gulp/tasks'
+$ = do require 'gulp-load-plugins'
 
-
-root_dir = './main'
-static_dir = "#{root_dir}/static"
-paths =
-  clean: [
-      "#{static_dir}/dst"
-      "#{static_dir}/ext"
-      "#{static_dir}/min"
-    ]
-  watch: [
-      "#{static_dir}/**/*.css"
-      "#{static_dir}/**/*.js"
-      "#{root_dir}/**/*.html"
-      "#{root_dir}/**/*.py"
-    ]
-
-
-run = (option) ->
-  proc = exec "python -u run.py -#{option}"
-  proc.stderr.on 'data', (data) -> process.stderr.write data
-  proc.stdout.on 'data', (data) -> process.stdout.write data
-
-
-gulp.task 'clean', ->
-  del paths.clean
-
-gulp.task 'bower_install', ->
-  $.bower()
-
-gulp.task 'ext', ['bower_install'], ->
-  gulp.src(main_bower_files(),
-    base: 'bower_components'
-  ).pipe gulp.dest("#{static_dir}/ext")
-
-gulp.task 'reload', ->
-  $.livereload.listen()
-  gulp.watch(paths.watch).on 'change', $.livereload.changed
-
-gulp.task 'watch', ->
-  run 'w'
-
-gulp.task 'run', ->
-  argv = process.argv.slice(2)
-  argv_lenght = Object.keys(argv).length
-  if argv_lenght <= 1
-    run 's'
-  else
-    known_options =
-      default:
-        C: false
-        c: false
-        h: false
-        m: false
-        s: false
-        w: false
-    options = minimist argv, known_options
-    for k of known_options.default
-      if options[k]
-        run k
-        break
-
-gulp.task 'default', ['reload', 'run', 'watch']
+gulp.task 'default',
+  'Start the dev_appserver.py, watching changes and LiveReload.
+  Available options - please see "run" task.',
+  $.sequence 'run', ['watch', 'reload']
--- a/run.py
+++ b/run.py
@@ -10,12 +10,10 @@ import platform
 import shutil
 import socket
 import sys
-import time
 import urllib
 import urllib2
 
 import main
-from main import config
 
 
 ###############################################################################
@@ -23,21 +21,8 @@ from main import config
 ###############################################################################
 PARSER = argparse.ArgumentParser()
 PARSER.add_argument(
-    '-w', '--watch', dest='watch', action='store_true',
-    help='watch files for changes when running the development web server',
-  )
-PARSER.add_argument(
-    '-c', '--clean', dest='clean', action='store_true',
-    help='recompiles files when running the development web server',
-  )
-PARSER.add_argument(
-    '-C', '--clean-all', dest='clean_all', action='store_true',
-    help='''Cleans all the pip, Node & Bower related tools / libraries and
-    updates them to their latest versions''',
-  )
-PARSER.add_argument(
-    '-m', '--minify', dest='minify', action='store_true',
-    help='compiles files into minified version before deploying'
+    '-d', '--dependencies', dest='install_dependencies', action='store_true',
+    help='install virtualenv and python dependencies',
   )
 PARSER.add_argument(
     '-s', '--start', dest='start', action='store_true',
@@ -52,12 +37,12 @@ PARSER.add_argument(
     help='the port to start the dev_appserver.py',
   )
 PARSER.add_argument(
-    '-f', '--flush', dest='flush', action='store_true',
-    help='clears the datastore, blobstore, etc',
-  )
-PARSER.add_argument(
     '--appserver-args', dest='args', nargs=argparse.REMAINDER, default=[],
     help='all following args are passed to dev_appserver.py',
+  )
+PARSER.add_argument(
+    '-v', '--version', dest='show_version', action='store_true',
+    help='Show gae-init version',
   )
 ARGS = PARSER.parse_args()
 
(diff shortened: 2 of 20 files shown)


topless:
Very legit, I have something similar but I didn't go thoroughly through it.

gulpfile.coffee
```diff
@@ -1,69 +1,217 @@
...
+    ]
+  clean_py: [
+      dir_lib
+      dir_venv
+    ]
+  style: [
+      "#{dir_style}/style.less"
+    ]
+  script: [
+      "#{dir_script}/**/*.coffee"
```
lipis: That is quite nice.. as you don't want to change anything because usually it will work..

lipis:
Very legit indeed..! Of course if you want to change or split the external in more files.. you'll have to change the .html file as well.. but in the end of the day is not that often that you have to do it.. and it's just a matter of documentation or realising what to do in that case..! I'll run couple of tests and we have to check Windows anyway and I don't see any reason not to go for it :D :dancer:

lipis:
@topless if you can test it as well and give feedback would be nice..

lipis:
@gmist in case you didn't know this one is pretty cool: https://github.com/tjunnone/npm-check-updates

gmist:
Interesting tool, thanks

gmist:
Now it working on Windows
screen shot 2015-03-04 at 7 58 00 pm

lipis:
:dancers: :100:

run.py
```diff
@@ -23,21 +21,8 @@
...
-  )
-PARSER.add_argument(
-    '-C', '--clean-all', dest='clean_all', action='store_true',
-    help='''Cleans all the pip, Node & Bower related tools / libraries and
-    updates them to their latest versions''',
-  )
-PARSER.add_argument(
-    '-m', '--minify', dest='minify', action='store_true',
-    help='compiles files into minified version before deploying'
+    '-v', '--venv', dest='install_venv', action='store_true',
```
lipis: `-v` is usually for version.. maybe something like:

```
    '-d', '--dependencies', dest='install_dependencies', action='store_true',
```

gulpfile.coffee
```diff
@@ -1,69 +1,214 @@
...
+
+gulp.task 'clean',
+'Clean the project from temp files, python compiled files
+and minified styles and scripts.', ->
+  del paths.clean
+  del paths.clean_py
 
 
-gulp.task 'clean', ->
+gulp.task 'clean_all',
```
lipis: instead of `clean_all` what about something like `init`?

because basically that's what it does..

gulpfile.coffee
```diff
@@ -1,69 +1,214 @@
...
+    .pipe(gulp.dest("#{dir_min}/style"))
+    .pipe($.rename('style.css'))
+    .pipe($.minifyCss())
+    .pipe(gulp.dest("#{dir_min}/style"))
+
+
+gulp.task 'build',
+"Compiles styles & scripts files into minified version
+and pack python dependencies into #{file_lib}.",
+  $.sequence('clean', 'npm', 'pip', 'ext_install', 'ext', 'pip', ['script', 'style', 'make_lib_zip'])
```
lipis: `pip` is twice here.. is that intentional?

gulpfile.coffee
```diff
@@ -1,69 +1,214 @@
...
+
+gulp.task 'npm', false, ->
+  gulp.src('package.json').pipe($.start())
 
-gulp.task 'bower_install', ->
-  $.bower()
 
-gulp.task 'ext', ['bower_install'], ->
+gulp.task 'pip', false, ->
+  gulp.src('run.py').pipe($.start([{match: /run.py$/, cmd: 'python run.py -v'}]))
```
lipis: Parenthesis could be avoided :) 

```coffeescript
gulp.src('run.py').pipe $.start [{match: /run.py$/, cmd: 'python run.py -v'}]
```

gulpfile.coffee
```diff
@@ -1,69 +1,214 @@
...
+gulp.task 'bower', false, ->
+  if /^win/.test process.platform
+    start_map = [{match: /bower.json$/, cmd: 'node_modules\\.bin\\bower install'}]
+  else
+    start_map = [{match: /bower.json$/, cmd: 'node_modules/.bin/bower install'}]
+  gulp.src('bower.json').pipe($.start(start_map))
+
+
+gulp.task 'npm', false, ->
+  gulp.src('package.json').pipe($.start())
```
lipis:
```coffeescript
 gulp.src('package.json').pipe $.start()
```

gulpfile.coffee
gmist: mistake

lipis:
It looks great and it works as expected when everything is installed..! But the behaviour before was a bit different. The good thing was that if you would make any changes to `bower.json`, `package.json` or `requirements.txt` no matter what you would choose to run, it would execute the bower install or npm install or pip when needed before compiling or actually running. So in a way it was a bit smarter and if something would go wrong or you wanted to update you would run the `-C` as a last resort.

That's why the `*.guard` files were there and this is still a case for pip related stuff, but not for `bower` and `npm` anymore. What about leaving the installations parts as part of the run.py doctor's checks and they would take care of the fact is bower install or npm install should run again (every time gulp is being called).

Does that make any sense?

gmist:
I meaningly removed guards section because time of execution of the relevant tasks are very few - approximately 2.5 seconds (assuming that the config files are not changed). If we leave the logic of the guards in ```run.py```, we lose about a 1 second (~500 ms for each task) to call Python from Gulp. This difference seemed to me not big enough to complicate logic of ```gulfile.coffee``` or leave this stuff in ```run.py```.
screen shot 2015-03-12 at 00 38 37
screen shot 2015-03-12 at 00 38 50

For example, time of checking ```pip``` guard
screen shot 2015-03-12 at 00 58 11

lipis:
Right I see your point.. and it's not that bad..!

What about then adding one task that is called `install` that has `npm, bower, pip` and call `install` first before the `default`?

The current `default` task is now taking around ~6 seconds.. so adding an extra of ~2 seconds (if nothing changed) is not that bad and you are safe.

lipis:
and maybe to [take it to 11](https://www.youtube.com/watch?v=KOO5S4vxi0o) we could add one more set of files to watch (bower.json, package.json, requirements.txt) and run the `install` task when these files are changed when gulp is already running.. wdyt?

gmist:
+1 good idea.

gulpfile.coffee
```diff
@@ -1,69 +1,239 @@
...
+      .pipe gulp.dest "#{dir_min}/script"
+
+
+gulp.task 'script_dev', false, ->
+  gulp.src paths.script
+      .pipe do $.sourcemaps.init
+      .pipe $.coffee().on 'error', $.util.log
+      .pipe $.concat 'script.dev.js'
+      .pipe do $.sourcemaps.write
+      .pipe gulp.dest "#{dir_min}/script"
```
lipis: indent.. <<

lipis:
Pretty solid.. will test over the weekend :)

lipis:
It looks like if bower hangs for some reason (often does that)... the only way to reset now is to do:

```
gulp init
npm install
gulp
```

because `gulp init` deletes everything.. and by deleting `node_modules` you have to run the `npm install`... can think of anything else that either doing the above or doing that using `run.py`..

(the old `./run.py -C` would clean everything but also reinstall a clean version of them... that's what I had in mind with the `init` but obviously I didn't think of the consequences)

gmist:
```gulp rebuild``` provides the most closest behavior.

**Comment by lipis** [comment]
More of the times is the bower that hangs, but sometimes it could be anything on poor connections.. what about on top of just `gulp bower` we could have also `gulp bower reset` will clean only the Bower related stuff and reinstall only bower.. and then same for pip and if something is wrong with npm then `gulp init` and/or `npm install`

File: gulpfile.coffee [hunk 8]
```diff
@@ -1,69 +1,239 @@
...
+    .pipe $.concat 'script.js'
+    .pipe do $.uglify
+    .pipe gulp.dest "#{dir_min}/script"
+
+
+gulp.task 'script_dev', false, ->
+  gulp.src paths.script
+    .pipe do $.sourcemaps.init
+    .pipe $.coffee().on 'error', $.util.log
+    .pipe $.concat 'script.dev.js'
```
topless: The problem with this approach I faced in the past is as the JavaScript codebase grows the build starts to slow down due to the concat. I have used [gulp-inject](https://github.com/topless/gae-init/blob/angular/gulpfile.coffee), be aware that they have to be seperate async tasks in order to do all the injections (libs and scripts). There is also another plugin I think called gulp-newer which rebuilds only the changed file. Great work by the way +1!

joenorton:
+1
I had somehow gummed up my minification of JS with the last setup so I just made the transition to this new system, and once I sorted out some permissions issues on my end - it appears to be working great. 

Also very convenient to have gulp deploy in there!

Nice job guys.

gmist:
Form [Getting started](http://getbootstrap.com/getting-started/):

*If you're compiling Bootstrap from its Less/Sass source and not using our Gruntfile, you'll need to integrate Autoprefixer into your build process yourself.*

lipis:
@gmist Fair enough :) Thanks for the link..

lipis:
As I never tried it before.. do you know if it avoids duplication of the rules if already there?

gmist:
I also have not used it before, so not exactly say ... I look at it as a magic

lipis:
With my recent tests.. works very nicely.. few comments:

- It reacts nicely on `bower.json` and `package.json` file changes.. but nothing happens when changing the `requirements.txt`. It should run the `pip` installation.

- I noticed that when I'm changing the `bower.json` the first time.. everything works perfectly fine. I made another change it throws an error: https://gist.github.com/lipis/b2615f15fd30ed3445ed

gmist:
Hmm... interesting... a bit later I'll check it

lipis:
After running the `pip install` we need to run the clean task to update the `lib.zip`

gmist:
This is done when the task 'build' (or 'deploy') is executed.  But it does not make sense for livereloading and development environment. In addition to that if you do clean the ```main/lib``` directory when ```dev_appserver.py``` is running it will break the livereloading.

lipis:
Right.. maybe then just a new gulp task that does the archiving of the lib directory?! That would be fast and the lib will be there on next refresh... because the pip install task puts the lib in the main/lib folder without deleting it.. and you don't need to reload dev_appserver when that happens... WDYT?

gmist:
I just do not see any reason for this. Because we do not use ```lib.zp``` with ```dev_appserver.py```.

lipis:
You are absolutely right...! My bad.. I thought it was used locally as well.. but it's not :)
https://github.com/gae-init/gae-init/blob/master/main/appengine_config.py

topless:
I cannot pass arguments to `dev_appserver.py` through gulp and to `appcfg.py`. Especially appcfg is needed in order to override application and version.

lipis:
Lets just not do everything at the same time.. @topless this issue could be a separate one and not part of this PR..

lipis:
```
$ npm-check-updates 

"gulp-minify-css" can be updated from ~0 to ~1 (Installed: 0.5.1, Latest: 1.0.0)
"gulp-zip" can be updated from ~2 to ~3 (Installed: 2.0.3, Latest: 3.0.2)
```

lipis:
In transit now.. but I think it's time for a squash.. :) I'll try to merge it tonight or tomorrow.. v3.0 is coming :)

lipis:
@topless mentioned something that was broken.. care to elaborate? As this is one of the biggest changes since v1.0 :dancer:

lipis:
:rocket:

lipis:
Houston.. we have a problem.. :/

It looks like the `gulp run` doesn't work and no errors.. did it after cloning the master:

```
lipis at iLipis in ~/gae-init on master
$ gulp run
[15:08:55] Using gulpfile ~/gae-init/gulpfile.js
[15:08:55] Starting 'run'...
[15:08:55] Starting 'install_dependencies'...
[15:08:55] Starting 'npm'...
[15:08:55] Finished 'run' after 35 ms
[15:08:55]  $ npm install
[15:08:56] Finished 'npm' after 1.2 s
[15:08:56] Starting 'pip'...
[15:08:56]  $ python run.py -d
[15:08:57] Finished 'pip' after 518 ms
[15:08:57] Starting 'bower'...
[15:08:57]  $ node_modules/.bin/bower install
[15:08:58] Finished 'bower' after 719 ms
[15:08:58] Starting 'copy_bower_files'...
[15:08:58] Finished 'copy_bower_files' after 109 ms
[15:08:58] Finished 'install_dependencies' after 2.55 s
[15:08:58] Starting 'ext:dev'...
[15:08:58] Starting 'script:dev'...
[15:08:58] Starting 'style:dev'...
[15:09:00] Finished 'script:dev' after 1.86 s
[15:09:00] Finished 'ext:dev' after 1.91 s

lipis at iLipis in ~/gae-init on master
```

gmist:
houston_we_have_a_problem_small

lipis:
Does the current master work for you?

lipis:
1965_04_gene_kranz_nasa_photo_s-65-22203

gmist:
Nope, it is somehow not working...

lipis:
In the branch that I had your PR it still works.. very strange :/

lipis:
It's something with the styles... no error and it's not being generated.. if you change the style.less to 
```css
body {
  background-color: red;
}
```

it works.. still trying to figure out which lib is causing it, but either case it's bad that it doesn't throw an error or something..

lipis:
Found the cause.. removed it for now.. if we can't display a proper error to let the user know.. then we might exclude it completely https://github.com/gae-init/gae-init/commit/c8d178b040d9debf1b42fccfdff49300ef440522

gmist:
Strange but it also works if you remove line 4 ```@import "../../ext/nprogress/nprogress";``` from style.less

lipis:
I see.. if you only going to use that line, it works as well.. so something is happening here :/

lipis:
Works without the source maps as well.. maybe it's better to include the `autoprefixer` and remove the sourcemaps until fixed.. WDYT?

gmist:
Please try #348

lipis:
Never played with source maps and less files.. did you check if they actually work?! (Will play around with them over the weekend)

gmist:
It seems to be working
screen shot 2015-05-01 at 18 00 23
screen shot 2015-05-01 at 18 02 10

lipis:
awesome.. thanks man!!

PR #331, opened 2015-03-14
Added prefix for the api urls.

topless:


gmist:
+1

lipis:
Supporting other versions will be quite a challenge though with this model to be future proof... :) 

What about having something like that in the main.py:

```python
api_v1 = helpers.Api(app, prefix='/api/v1')
```

and then importing that thing from main inside the `api.v1` modules..

doing that it will be easier in the future to just create new `v2`, `v3` modules and change only the import and in the main.py it will look like like:

```python
api_v1 = helpers.Api(app, prefix='/api/v1')
api_v2 = helpers.Api(app, prefix='/api/v2')
api_v3 = helpers.Api(app, prefix='/api/v3')
...
```

WDYT?

main/main.py
```diff
@@ -22,7 +22,7 @@
 import task
 
 from api import helpers
-api = helpers.Api(app)
+api = helpers.Api(app, prefix='/api/v1')
```
lipis:
```python
api_v1 = helpers.Api(app, prefix='/api/v1')
```

topless:
Well in my head there are very few apis that went to version 2. Maybe `api1` or `api_1` or fuck it and leave it as api cause its 3 chars and really sweet. What @gmist @mdxs @joernhees you guys think?

lipis:
https://developers.google.com/analytics/devguides/collection/android/v4/
https://developers.google.com/maps/documentation/javascript/

:)  APIs are going up in general :P

topless:
I am rebasing it on top of the master and I was looking around for the `import api` in order to do the refactor. Considering your suggestion
`api_v1 = helpers.Api(app, prefix='/api/v1')`
`import api_v1.v1`
You see my point, so what if we refactored the api directory into `api_v1` and remove the `v1` so we can escape the v1.v1 import ugliness.

main/api/v1/config.py
```diff
@@ -10,7 +10,7 @@
 from main import api
```
lipis: this will become 
```python
from main import api_v1
```

topless:
I think I got confused but I was referring to the import in main.py line 27.

lipis:
yep.. that will remain as is :D

topless:
I got confused because the variable `api` and the module `api` share the same name .. although they are different things :)

PR #332, opened 2015-03-14
Fix import config

### LINKED ISSUE #330: /api/v1/config/ creates Internal Server Error.

gmist:
Closes #330


lipis:
Should be in all api files... noticed that in another project.. so just for the future maybe adding it to the user.py as well..?

gmist:
+1 (although it is not necessarily in this case)

PR #336, opened 2015-03-22
Removed SweetAlert

### LINKED ISSUE #269: Remove the SweetAlert
Maybe we should remove it after all.. it needs maintenance and bunch of other things, so I think we  were rushed a bit with this one.. WDYT?

lipis:
Fixes #269


lipis:
I just reverted to the old version.. until there will be a better way :)

PR #337, opened 2015-03-27
Flask urls available from JavaScript

topless:
With this approach, we define our paths only in one place, the server. You can access the endpoints you love so much with their names as you do with `{{flask.url_for()}}` when using Jinja 

Somewhere in your JavaScript
```
console.log(flask_util.url_for('api.user.list'))
console.log(flask_util.url_for('api.user', {user_key: "TheUserIdOrWhatever"}))
```

### DIFF
--- a/main/main.py
+++ b/main/main.py
@@ -26,6 +26,8 @@ api_v1 = helpers.Api(app, prefix='/api/v1')
 
 import api.v1
 
+from flask_util_js import FlaskUtilJs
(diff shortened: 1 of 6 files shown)


lipis:
This is quite an exposure though: https://1-dot-gae-init.appspot.com/flask_util.js 

Didn't have time to check much regarding possible minification or defining your own URL for it?

Interesting for sure.. just some concerns regarding privacy..

PR #340, opened 2015-04-03
Introducing build directory.

topless:
In order to make the build process more clean and flexible I have a couple of suggestions.

I would like to introduce build directory as the place which will accommodate the products of our pre processors (Coffee, Less). At this point we are trying to do everything at once. My proposal is to build what we have to build in the relevant dir, and when we want to deploy we minify and concat what we have there. Like this we also avoid re-writing tasks for dev and prod.

Previous ext directory was also moved inside the build dir.

This pr is not finished is just some ground work and proof of concepts.

Any suggestions, ideas?


lipis:
As this is based on another PR.. it's hard to see what is going on.. will close it for now and we'll see after the #321..

PR #341, opened 2015-04-15
Updated endpoints for OAuth

lipis:
Since we deprecated the `/_s/` maybe we should update these endpoints as well.. but I'll leave the old ones for backwards compatibility as this a bit pain to update :)

Maybe even to something smaller?
```
/api/callback/brand/auth/
```

any other ideas?


gmist:
Maybe
```
/api/auth/callback/brand/
```
?

lipis:
Yes much better and shorter..

lipis:
photo on 19-04-15 at 11 55 2

PR #342, opened 2015-04-15
Removed greetings and goodbyes

lipis:
I'm usually removing those in other projects.. WDYT?


dhstack:
I don't mind them, but I also don't think they're necessary

PR #343, opened 2015-04-22
Fix Google IP

gmist:


lipis:
Why did they disable it?! :( this might happen again I guess.. what about checking for www.google.com? Will take more time for the DNS resolution.. but might be more future proof

lipis:
:dancer: you did it..

run.py
```diff
@@ -124,7 +124,7 @@
 # Other global variables
 ###############################################################################
 CORE_VERSION_URL = 'https://gae-init.appspot.com/_s/version/'
-INERNET_TEST_URL = 'http://74.125.228.100'
+INERNET_TEST_URL = 'http://google.com'
```
lipis: as this is a redirect.. maybe it's better to use the https://www.google.com or it doesn't matter?

run.py
```diff
@@ -124,7 +124,7 @@
 # Other global variables
 ###############################################################################
 CORE_VERSION_URL = 'https://gae-init.appspot.com/_s/version/'
-INERNET_TEST_URL = 'http://74.125.228.100'
+INERNET_TEST_URL = 'http://www.google.com'
```
lipis: `httpS` :D

gmist: I do not think that this is important, but so be it.

lipis:
now you can see where I'm :) and probably that's why there is not much going on.. :( https://gae-init.appspot.com/stats/ stupid Google :)

gmist:
Hot place for holidays

joenorton:
:+1:  Ah, here I was thinking I was having wifi issues!

PR #348, opened 2015-05-01
Fixed incorrect working with sourcemaps

gmist:


lipis:
That fixes it.. so add the autoprefixer now and I'll merge it :)

PR #356, opened 2015-05-05
Added Reddit OAuth

lipis:
So for some reason when I set the `redirect uri` to `httpS` on Reddit's side.. it doesn't work.. It tries to redirect to `http` and it fails..

screen shot 2015-05-05 at 23 20 58

Here is how the URL looks like even though in the settings I have it defined with HTTPS :/
```
https://ssl.reddit.com/api/v1/authorize?response_type=code&client_id=XXXXX&redirect_uri=http%3A%2F%2Fgae-init.appspot.com%2Fapi%2Fauth%2Fcallback%2Freddit%2F&scope=identity&state=0b2f9941a58c4eeb8e2c01e69e2c3f61
```

### DIFF
--- a/main/auth/__init__.py
+++ b/main/auth/__init__.py
@@ -10,6 +10,7 @@ from .google import *
 from .instagram import *
 from .linkedin import *
 from .microsoft import *
+from .reddit import *
 from .twitter import *
 from .vk import *
 from .yahoo import *
--- a/main/auth/auth.py
+++ b/main/auth/auth.py
@@ -288,6 +288,7 @@ def urls_for_oauth(next_url):
       'instagram_signin_url': url_for_signin('instagram', next_url),
       'linkedin_signin_url': url_for_signin('linkedin', next_url),
       'microsoft_signin_url': url_for_signin('microsoft', next_url),
+      'reddit_signin_url': url_for_signin('reddit', next_url),
       'twitter_signin_url': url_for_signin('twitter', next_url),
       'vk_signin_url': url_for_signin('vk', next_url),
       'yahoo_signin_url': url_for_signin('yahoo', next_url),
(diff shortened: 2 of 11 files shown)


lipis:
So for some reason when I set the URL to HTTPS on Reddit's side.. it doesn't work.. It tries to redirect to http and it fails..

screen shot 2015-05-05 at 23 20 58

Here is how the URL looks like even though in the settings I have it defined with HTTPS :/
```
https://ssl.reddit.com/api/v1/authorize?response_type=code&client_id=XXXXX&redirect_uri=http%3A%2F%2Fgae-init.appspot.com%2Fapi%2Fauth%2Fcallback%2Freddit%2F&scope=identity&state=0b2f9941a58c4eeb8e2c01e69e2c3f61
```

lipis:
Is it the same on your side @gmist?

gmist:
At first view I have the same problem, tomorrow I'll check it attentively.

lipis:
Thanks, I filed a bug report anyway :) http://www.reddit.com/r/bugs/comments/34zlne/when_i_set_the_redirect_uri_to_https_in_my_reddit/

lipis:
I _smells_ like it should be here: https://github.com/lipis/gae-init/blob/reddit-oauth/main/auth/reddit.py#L36

gmist:
In ```flask_oauthlib/client.py``` line 518
``` python
def authorize(self, callback=None, state=None, **kwargs):
  ...
  session['%s_oauthredir' % self.name] = callback
  ...
```
so I think that the problem is not here

lipis:
but on Reddit's side?

gmist:
No, but I think I found the problem...

lipis:
Messed it up with the squashes mashes lashes.. but I think now I know :)

p89inhk

lipis:
Live: https://gae-init.appspot.com/signin/

PR #359, opened 2015-05-09
Correct class for the header's container

lipis:
Because otherwise it looks like this when is not fixed..

screen shot 2015-05-09 at 23 09 32


lipis:
Never mind.. :dancer:

PR #360, opened 2015-05-11
Added Bootswatch

lipis:
#### https://bootswatch.com/

### DIFF
--- a/main/static/src/style/style.less
+++ b/main/static/src/style/style.less
@@ -3,6 +3,9 @@
 @import "../../ext/lesshat/build/lesshat-prefixed";
 @import "../../ext/nprogress/nprogress";
 
+@import "../../ext/bootswatch/flatly/variables";
+@import "../../ext/bootswatch/flatly/bootswatch";
+
 @import "base";
 @import "variables";
 @import "test";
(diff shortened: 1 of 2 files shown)


gmist:
+1

main/static/src/style/style.less
```diff
@@ -3,6 +3,9 @@
 @import "../../ext/lesshat/build/lesshat-prefixed";
 @import "../../ext/nprogress/nprogress";
 
+@import "../../ext/bootswatch/flatly/variables";
+@import "../../ext/bootswatch/flatly/bootswatch";
+
```
dhstack: @lipis 
What if you did something like this?
```
@theme: "flatly";
@import "themes/@{theme}/variables.less";
@import "themes/@{theme}/bootswatch.less";
```

Are you planning on having a way to choose one of the bootswatch themes from the admin area?

lipis: Use multicursor/multiselect (Cmd+D on Sublime) and you're set.. :)

lipis: All the things tagged as `example` won't be merged so I'm not planning to do anything more fancy, especially with the admin area as it will require building 10+ different CSS and in real life situations you have only one theme that is none of the these ones.. (at least you shouldn't launch a product with a default theme)

P.S. As these are the examples eventually I'll clean up the comments.. take no offence..

dhstack: Gotcha, missed the `example` tag

PR #365, opened 2015-05-14
Removed default http for callbacks and deprecated URLs

lipis:
It was added because of Dropbox that asks only for HTTPS (which is a good thing) but it breaks other stuff when the default is `http`. Had the same problem with Google and others.. it should be OK with this change

### DIFF
--- a/main/auth/auth.py
+++ b/main/auth/auth.py
@@ -319,7 +319,7 @@ def save_request_params():
     }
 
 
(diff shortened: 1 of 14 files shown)


lipis:
Actually it also doesn't make sense to keep the old `/_s/callback/` URLs as in my mind I though that the callback URLs are set by the providers, but this is not true. So that change already broke a lot of things :D

PR #372, opened 2015-05-27
Removed source maps from ext libraries.

topless:
Source maps for libs caused great delay when developing.


lipis:
Interesting.. anything against it?

gmist:
With them much easier to debug code if an error occurs in the external library.

lipis:
true that.. and maybe it's better to keep it here and if for anyone is too slow.. because there are too many external libs.. they should already know how to disable source maps :)

PR #373, opened 2015-05-27
Reloading only when the necessary files has build.

topless:
The files that should trigger the reload when changed are not the whole, static folder but the dev folder in it.

### DIFF
--- a/gulp/tasks/watch.coffee
+++ b/gulp/tasks/watch.coffee
@@ -6,8 +6,8 @@ paths = require '../paths'
 gulp.task 'reload', false, ->
   do $.livereload.listen
   gulp.watch([
-    "#{paths.static.root}/**/*.css"
-    "#{paths.static.root}/**/*.js"
+    "#{paths.static.root}/dev/style/*.css"
+    "#{paths.static.root}/dev/script/*.js"
     "#{paths.main}/**/*.html"
     "#{paths.main}/**/*.py"
   ]).on 'change', $.livereload.changed
--- a/main/__init__.py
+++ b/main/__init__.py
@@ -13,4 +13,4 @@ License MIT, see LICENSE for more details.
 
 """
 
-__version__ = '3.1.1'
+__version__ = '3.2.0'


gmist:
what will happen if we will update/add any ```ext``` library?

lipis:
That's a good point @gmist, but I think it's not happening that often and when there are many external libs @topless argument is that it's getting very slow or even throwing errors, that there are too many files to watch.. is that true?

lipis:
@gmist actually this task is only for reloading which doesn't make sense to keep track of `/ext/` folder as the generated stuff from it endup in `/dev/` anyway through https://github.com/gae-init/gae-init/blob/3.2.0/gulp/tasks/watch.coffee#L25

so it totally makes sense..

gulp/tasks/watch.coffee
```diff
@@ -6,8 +6,8 @@ paths = require '../paths'
 gulp.task 'reload', false, ->
   do $.livereload.listen
   gulp.watch([
-    "#{paths.static.root}/**/*.css"
-    "#{paths.static.root}/**/*.js"
+    "#{paths.static.root}/dev/style/*.css"
+    "#{paths.static.root}/dev/script/*.js"
```
lipis: Or just `"#{paths.static.dev}/**/*.{css,js}"` for both

gulp/tasks/watch.coffee
```diff
@@ -6,8 +6,8 @@ paths = require '../paths'
...
     "#{paths.main}/**/*.html"
     "#{paths.main}/**/*.py"
```
lipis: maybe the same here: `"#{paths.main}/**/*.{html,py}"`

topless:
@gmist, @lipis in my head watching everything was too "expensive" for something that doesn't happen that often, like adding a lib. As for the watch.coffee#L25, I believe it can be removed.

lipis:
@topless it was a good catch.. as we don't have to reload the app if `ext` folder will change.. but only when the generated `ext.js` will be created.. as this is the one that is being used..

PR #379, opened 2015-06-04
Exclude .so shared libraries

gmist:
From packages like [mistune](https://pypi.python.org/pypi/mistune/) because they can cause problems with the import.

### DIFF
--- a/run.py
+++ b/run.py
@@ -171,7 +171,7 @@ def install_py_libs():
 
   exec_pip_commands('pip install -q -r %s' % FILE_REQUIREMENTS)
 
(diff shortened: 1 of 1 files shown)


lipis:
So mistune is better than #366? Claims to be faster.. :)

gmist:
[jinja2-markdown](https://github.com/danielchatfield/jinja2_markdown/blob/master/setup.py) based on Markdown package
``` python
	install_requires=[
		'Markdown>=2.3.1',
		'Jinja2>=2.4'
	]
```
so mistune [will be faster](http://lepture.com/en/2014/markdown-parsers-in-python).

PR #383, opened 2015-06-16
Fix repository URL

gmist:
"licenses" property [is deprecated](https://docs.npmjs.com/files/package.json#license)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,15 +1,10 @@
 {
   "name": "gae-init",
   "author": "Panayiotis Lipiridis <lipiridis@gmail.com>",
-  "licenses": [
-    {
-      "type": "MIT",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,15 +1,10 @@
...
-      "type": "MIT",
-      "url": "http://opensource.org/licenses/MIT"
-    }
-  ],
-  "repository": {
+  "license": "MIT",
+   "repository": {
     "type": "git",
-    "url": "http://github.com/gae-init"
+    "url": "https://github.com/gae-init/gae-init"
```
lipis: `https` yes.. but as it has other flavours and I didn't want to update them individually as it will produce conflicts I wanted to have the parent project..

PR #395, opened 2015-07-13
'gcloud preview app' temporary workaround

gmist:
It is a temporary fix - ugly code but seems to work

### DIFF
--- a/gulp/tasks/build.coffee
+++ b/gulp/tasks/build.coffee
@@ -25,7 +25,7 @@ gulp.task 'deploy', 'Deploy project to Google App Engine.', ['build'], ->
     options_str += if k.length > 1 then " --#{k} #{options[k]}" else " -#{k} #{options[k]}"
 
   gulp.src('run.py').pipe $.start [
-      {match: /run.py$/, cmd: "appcfg.py update main #{options_str}"}
+      {match: /run.py$/, cmd: "python run.py --deploy"}
     ]
 
 
--- a/run.py
+++ b/run.py
@@ -29,6 +29,10 @@ PARSER.add_argument(
     help='starts the dev_appserver.py with storage_path pointing to temp',
   )
 PARSER.add_argument(
+  '--deploy', dest='deploy', action='store_true',
+  help=' deploy your app to Google App Engine',
+)
+PARSER.add_argument(
     '-o', '--host', dest='host', action='store', default='127.0.0.1',
     help='the host to start the dev_appserver.py',
   )
@@ -52,6 +56,7 @@ ARGS = PARSER.parse_args()
 ###############################################################################
 BAD_ENDINGS = ['pyc', 'pyo', '~']
 GAE_PATH = ''
+GAE_CMD = ''
 IS_WINDOWS = platform.system() == 'Windows'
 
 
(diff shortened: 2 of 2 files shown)


gmist:
Perhaps we should rewrite 'gulp deploy' command because the properties 'application' and 'version' is deprecated now and we need to set them manually when we exec 'gulp deploy' command... something like that:
```shell
gulp deploy -v 20150713
```
```
gulp deploy_default -v 1
```

gmist:
``` shell
gulp deploy -v 20150713
```
will be converted to
``` shell
gcloud preview app deploy --version 20150713 main/app.yaml
```
and
```
gulp deploy_default -v 1
```
converted to
```
gcloud preview app deploy --set-default --version 20150713 main/app.yaml
```

lipis:
Yes, there are quite a few things that are done totally different :/ The `application` can be replaced with `project` (`gcloud help` to get the global flags), but overall I have to admit that I'm not entirely happy on how things done with the new command.. that is currently under beta actually.

nazroll:
funny thing about their "beta"... it's a permanent change. sigh.

lipis:
From [Andrew Jessup](https://plus.google.com/+AndrewJessup) @ajessup

Hey folks - apologies for the confusion here. Long story short, prior to last week's announcement we bundled appcfg.py and dev_appserver.py in two places:

1) our stand alone GAE SDKs at https://cloud.google.com/appengine/downloads as well as 2) in the Cloud SDK (which also has the gcloud preview app * command group).

Unfortunately the versions of appcfg.py and dev_appserver.py of the SDKs bundled with the Cloud SDK weren't great. Not only were they released later than the"official" App Engine SDKs but they frequently had regressions and other bugs that we weren't catching. We also found that the vast majority of customers were using the stand-alone SDKs anyway.

As such we decided it would be best to simply direct users to the official App Engine SDKs for the time being rather than copy them into the Cloud SDK. That way users are using SDKs that are both more up to date and better tested.

So, for now - the App Engine SDKs https://cloud.google.com/appengine/downloads are the ones we officially recommend customers use (an exception to this is Managed VMs, which are also currently in Beta).

We are of course, in the meantime, working hard on getting gcloud app out of preview and being fully supported - long term we are working to make this the default tool for all customers of App Engine. As you and others have noted though - we've got a bit of work to do before we're ready to remove the Preview label.

Again, sorry for the confusion here folks. Let me know if there's anything else I can help clarify.

gmist:
It PR works with the latest version of the GAE SDK and the latest version of the Cloud SDK, but we will have to keep track of its operability after each update Cloud SDK or we should temporarily abandon the use of the Cloud SDK.

lipis:
Maybe it makes to clean this PR from the GAE SDK and leave only the potential solution for current version of Cloud SDK for those who wants to use it.. will just merge this.. and merge it when it is out of preview.. WDYT?

gmist:
I don't even know what to say... it may be too difficult for most people who are used Cloud SDK... maybe

lipis:
What do you mean more difficult?!

gmist:
It will require additional actions (like merge/use other remote branch/install of the GAE) for users who use the Cloud SDK. I like when everything works out of the box, but maybe you're right and is better to make it as a separate branch until Cloud SDK is not released.

lipis:
I think it's going to be cleaner that way.. and if anything we could actually merge it earlier if we see that it goes to that direction or something! Plus if someone is going to have issues we could refer to that PR here..

lipis:
I think we can simply go with this #452

PR #399, opened 2015-07-20
Allow to get maximum results from query

gmist:
```
limit
Maximum number of results to return. If set to None, all available results will be retrieved.
```
[from here](https://cloud.google.com/appengine/docs/python/ndb/queryclass#kwdargs_options)

Example:
``` python
  user_dbs, _ = model.User.get_dbs(email=util.param('email'), limit=None)
```
It returns all available users from the DB.

### DIFF
--- a/main/util.py
+++ b/main/util.py
@@ -56,9 +56,8 @@ def get_next_url(next_url=''):
 # Model manipulations
 ###############################################################################
 def get_dbs(
-    query, order=None, limit=None, cursor=None, keys_only=None, **filters
+    query, order=None, limit=0, cursor=None, keys_only=None, **filters
   ):
-  limit = limit or config.DEFAULT_DB_LIMIT
   cursor = Cursor.from_websafe_string(cursor) if cursor else None
   model_class = ndb.Model._kind_map[query.kind]
   if order:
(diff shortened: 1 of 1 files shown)


gmist:
I'm sorry, but it breaks the logic of the function util.param()... Is better to roll back to version 3a677f038e432737e7a59c3896c55af1d3b3da95

gmist:
Maybe something like [that](https://github.com/gmist/five-studio2/commit/09b9b8986e7f2e002aaea97c6520bf34a2fcbb55#diff-873b25319843fdb5d75c9fad811dcbeeR77)

and somewhere in a template
``` html
        <button type="button" class="btn btn-primary" disabled>Limit</button>
        {{utils.filter_by_link('limit', 16)}}
        {{utils.filter_by_link('limit', 64)}}
        {{utils.filter_by_link('limit', -1, label='All')}}
```

lipis:
No problem.. rolled it back for now.. will check tomorrow for the alternatives :) cheers..

lipis:
@gmist yes something like that should be good to go.. Like a little easter egg with the `-1`

lipis:
@gmist Create a new PR if you like.. :)

PR #403, opened 2015-07-23
Next/prev cursors refactoring

gmist:
I think it also should be compatible with existing code (with gae-init v3.2.0). WDYT?

### DIFF
--- a/main/api/helpers.py
+++ b/main/api/helpers.py
@@ -35,7 +35,7 @@ def handle_error(e):
     }), e.code
 
 
-def make_response(data, marshal_table, next_cursor=None, prev_cursor=None):
+def make_response(data, marshal_table, cursors=None):
   if util.is_iterable(data):
     response = {
         'status': 'success',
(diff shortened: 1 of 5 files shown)


lipis:
I like it :) if it works out of the box with the old one.. will be nice..

lipis:
screen shot 2015-07-23 at 22 25 41

gmist:
I'm just a little bit experimenting with the types and forgot to change them. Now it should work.

main/api/helpers.py
```diff
@@ -35,7 +35,7 @@ def handle_error(e):
     }), e.code
 
 
-def make_response(data, marshal_table, next_cursor=None, prev_cursor=None):
+def make_response(data, marshal_table, next_cursor=None):
```
lipis: just `cursors` here is enough

gmist: I leave it for backwards compatibility, if someone is using named parameters
``` python
make_response(data, table, next_cursor=my_cool_cursor)
```

lipis: I really doubt that with only three params and one optional that would be the case :)

main/api/helpers.py
```diff
@@ -44,11 +44,15 @@ def make_response(data, marshal_table, next_cursor=None, prev_cursor=None):
         'result': map(lambda l: restful.marshal(l, marshal_table), data),
       }
     if next_cursor:
-      response['next_cursor'] = next_cursor
-      response['next_url'] = util.generate_next_url(next_cursor)
-    if prev_cursor:
-      response['prev_cursor'] = prev_cursor
-      response['prev_url'] = util.generate_next_url(prev_cursor)
+      if isinstance(next_cursor, dict):
+        response['next_cursor'] = next_cursor['next_cursor']
```
lipis: then maybe `cursors['next']` and `cursors['prev']`

gmist: +1 is a good idea

lipis:
Something is wrong.. with the previous link.. to reproduce locally:

- Create 5 users
- [localhost:8080/admin/user/?limit=2](http://localhost:8080/admin/user/?limit=2)
- Click next
- Click prev

Shows different results and sometimes (if you press around) no results.. 

(not sure if it's part of this PR or the previous one

gmist:
Hm... at the same time this problem does not reproduced if you use ordering, right?
[http://localhost:8080/admin/user/?order=-modified&limit=2](http://localhost:8080/admin/user/?order=-modified&limit=2)

lipis:
yes.. that's why I haven't noticed it when testing the other day.. but when tested on something else realised it :)

lipis:
Update the online version.. https://gae-init.appspot.com/user/?limit=4

PR #404, opened 2015-08-02
Vertically center table cell contents

dhstack:
On table cells with multiple lines, the rows weren't vertically aligned to the middle - now they are.


lipis:
Sorry for my late reply.. but I want to keep the custom styles to the minimum.. will have to fold on this one.

dhstack:
Noted

PR #423, opened 2015-10-13
Reformat

lipis:
Some people might kill me for that.. cause they asked for it in the past :) and maybe it's not a good idea as it might break bunch of stuff... but here is an idea :)


gmist:
tumblr_inline_nox99at6wg1raprkq_500

lipis:
giphy

lipis:
Maybe a really bad idea.. it is still just an idea.. so it won't be merged without a discussion :)

topless:
Fuck Yeah

I 've been busting Lipis ball's for this for years.. so pep linter gets happy [E126](http://pep8.readthedocs.org/en/latest/intro.html#error-codes).

lipis:
quite a few people were not happy about that one :) But still not sure if that is proper thing to do now... it will break bunch of things if you are upgrading.. maybe changing the version to 4 and have some instructions on how to avoid many conflicts?! (Like autoformat with PyCharm before merging)

gmist:
I think that switching to 4 (with Bootstrap 4 I guess) will be painful in any case, so reformat will not be the biggest problem.

lipis:
Was thinking of doing it before the Bootstrap 4.. it might take a while :) 

(Also regarding Bootstrap 4.. we might want to keep a branch for some support for Bootstrap 3.. it's quite a thing to switch to SASS for older projects)

lipis:
So I might go ahead with this one before Bootstrap 4... any strong objections?! It's for a greater good :dancer:

mdxs:
Do check with pep8 tool and see which ones you really want to suppress given your beliefs, using a `tox.ini` like the following could help:

```
[pep8]
ignore = E111,E114
max-line-length = 160
```

Some of the current suggestions might hit E302

lipis:
Am I forgetting something? Anything to add? I might merge this and release version 4.0.0 this week..

@gmist @topless @mdxs

lipis:
and may your gods be with ya! :fireworks: :godmode:

PR #427, opened 2015-10-20
Fixed bump counter call by using named parameter

mdxs:
The definition of the cache bump counter in `main/cache.py` reads `def bump_counter(key, time=3600, limit=4)` (note the order of named parameters after key), making a call like `bump_counter(k, 4)` use `time=4` (and default `limit=4`).

By making use of a named parameter inside `bump_auth_attempt` in `main/cache.py`, the intent is made explicit and this fixes in my opinion a subtle bug. Please note that I didn't test the result, but I would expect that the `form_with_recaptcha` in `main/auth/auth.py` would now start showing a Captcha when there have been 8 signin forms failed/requested within an hour (3600 sec) from the same IP.

Without the bug fixed, the Captcha might only have been shown when the user was able to fail the signin form 8 times in 4 sec?

### DIFF
--- a/main/cache.py
+++ b/main/cache.py
@@ -38,4 +38,4 @@ def get_auth_attempt():
 
 
 def bump_auth_attempt():
-  bump_counter(get_auth_attempt_key(), config.SIGNIN_RETRY_LIMIT)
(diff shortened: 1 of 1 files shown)


lipis:
Good one.. or changing the order of the arguments.. :)

PR #428, opened 2015-10-21
OAuth API for REST

lipis:
Here is an example of this API using cURL on `localhost`

```bash
# Using username/password you can store the cookie using form data, JSON or url params
curl -d 'username=test&password=123456' --cookie-jar cookie.txt http://localhost:8080/api/v1/auth/signin/
curl -d '{"username": "test", "password": "123456"}' -H 'Content-Type: application/json' --cookie-jar cookie.txt http://localhost:8080/api/v1/auth/signin/

# Using the cookie/session you can retrieve data
curl --cookie cookie.txt http://localhost:8080/api/v1/user/
curl --header 'Cookie:session=.eJxVj...nqlVs' http://localhost:8080/api/v1/user/
```

### DIFF
--- a/main/api/v1/__init__.py
+++ b/main/api/v1/__init__.py
@@ -1,4 +1,5 @@
 # coding: utf-8
 
+from .auth import *
 from .config import *
 from .user import *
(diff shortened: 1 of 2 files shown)


lipis:
You can see it in action in: https://github.com/lipis/timelapse/blob/master/pss-upload.py

PR #430, opened 2015-11-19
Middleware wrapper for appstats

topless:
I noticed today that appstats don't work any longer. You can verify by visiting /_ah/stats, despite the fact they are enabled in app.yaml we need this request wrapper.


topless:
You can read more about the appstats event recorder at the official [docs]. (https://cloud.google.com/appengine/docs/python/tools/appstats#EventRecorders)

PR #433, opened 2015-11-25
Removed NProgress

lipis:
I think I'll add this as an example PR instead.


mdxs:
There was no use of NProgress in plain gae-init?

lipis:
nope :) But I'll include it as PR example soonish.. if anyone wants to use it.. :)

PR #452, opened 2015-12-19
gcloud preview app deploy

### LINKED ISSUE #424: gcloud preview app deploy
From the https://dl.google.com/dl/cloudsdk/release/RELEASE_NOTES

> The App Engine Admin API is now the default path for deployments in the
`gcloud preview app deploy` command. This transition should be transparent,
but if you experience any issues and would like to revert to the old
behavior, please run `gcloud config set app/use_appengine_api false`.

lipis:
Fixes #424

### DIFF
--- a/gulp/tasks/build.coffee
+++ b/gulp/tasks/build.coffee
@@ -20,7 +20,10 @@ gulp.task 'deploy', 'Deploy project to Google App Engine.', ['build'], ->
   delete options['_']
   options_str = ''
   for k of options
-    if options[k] in [true, false]
+    if options[k] == false
+      options_str += " --no-#{k}"
+      continue
+    if options[k] == true
       options[k] = ''
     options_str += " #{if k.length > 1 then '-' else ''}-#{k} #{options[k]}"
 
--- a/gulp/tasks/build.coffee
+++ b/gulp/tasks/build.coffee
@@ -18,14 +18,14 @@ gulp.task 'rebuild',
 gulp.task 'deploy', 'Deploy project to Google App Engine.', ['build'], ->
   options = minimist process.argv
   delete options['_']
-  options_str = '--skip_sdk_update_check'
+  options_str = ''
   for k of options
     if options[k] == true
       options[k] = ''
     options_str += if k.length > 1 then " --#{k} #{options[k]}" else " -#{k} #{options[k]}"
 
   gulp.src('run.py').pipe $.start [
-      {match: /run.py$/, cmd: "appcfg.py update main #{options_str}"}
+      {match: /run.py$/, cmd: "gcloud preview app deploy main/app.yaml #{options_str}"}
     ]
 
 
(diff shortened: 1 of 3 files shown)


lipis:
```
gulp deploy
gulp deploy --project=foo
gulp deploy --project=foo --version=bar
etc..
```

gmist:
This will not work with GAE SDK, only with Cloud SDK?

lipis:
But is that a big issue? 

The gcloud is complaining about older versions being in PATH.. plus it will come sooner or later anyway..

gmist:
I think it's better to support them both... until GAE SDK will not be finally deprecated. Plus, GAE SDK is easier to use, because its installation is done with just one command:
```
brew install app-engine-python
```

lipis:
True the installation is easier and it's done once.. for the new comers not many things will change since most likely they will use the wrapper gulp command anyway.. and if anything we can simply improve this part http://docs.gae-init.appspot.com/requirement/#gae because Python tools are not coming out of the box anymore...

```bash
curl https://sdk.cloud.google.com | bash
gcloud components install app-engine-python
```

I agree that it really sucks that they are removing the application/version but there are some reasons behind it and I think in the long run it would be a nice thing to comply with it sooner than later :)

Also I'll change to version 4.0 so things will break :)

lipis:
Unless we'll have a new branch called `v4` (or something) that will have all of these changes.. I just wanted to change the version before the year end :dancer: :fireworks:

lipis:
I think it's time. WDYT? @gmist @topless

gmist:
Do it, comrade.

gulp/tasks/build.coffee
```diff
@@ -18,15 +18,16 @@ gulp.task 'rebuild',
 gulp.task 'deploy', 'Deploy project to Google App Engine.', ['build'], ->
   options = minimist process.argv
```
lipis: there are some issues with the [minimist](https://github.com/substack/minimist/issues/54) and in our case when the `--no-promote` is being used.. it is translated to `promote: false` :/

gulp/tasks/build.coffee
```diff
@@ -18,15 +18,19 @@ gulp.task 'rebuild',
...
   delete options['_']
-  options_str = '--skip_sdk_update_check'
+  options_str = ''
   for k of options
+    if options[k] == false
+      options_str += " --no-#{k}"
+      continue
```
lipis: spacial case for the `--no-promote` :(

gulp/tasks/build.coffee
```diff
@@ -18,15 +18,19 @@ gulp.task 'rebuild',
...
     if options[k] == true
       options[k] = ''
-    options_str += if k.length > 1 then " --#{k} #{options[k]}" else " -#{k} #{options[k]}"
+    options_str += " #{if k.length > 1 then '-' else ''}-#{k} #{options[k]}"
```
lipis: to reduce the line size..

lipis:
All righty then... :/ May your gods be with ya!

PR #455, opened 2015-12-21
Explicit define module name and instance class.

topless:


main/app.yaml
```diff
@@ -1,4 +1,6 @@
 application: gae-init
+module: default
+instance_class: F1
```
lipis: for less conflicts put it right after version

PR #458, opened 2016-01-07
Use gulp-cssnano instead of gulp-minify-css

gmist:
gulp-minify-css is now [deprecated](https://www.npmjs.com/package/gulp-minify-css).


gulp/tasks/style.coffee
```diff
@@ -9,8 +9,8 @@ gulp.task 'style', false, ->
   gulp.src config.style
   .pipe $.plumber errorHandler: util.onError
   .pipe do $.less
-  .pipe $.autoprefixer {cascade: false}
```
lipis: Is the autoprefixer included or being called by the cssnano or something?

gmist: `autoprefixer` is enabled by [default](http://cssnano.co/options/).

PR #454, opened 2015-12-20
Rewrite param() function via webargs

gmist:

PR #459, opened 2016-01-07
Explicitly check for method.

topless:
When I am trying to post a json array in the api, I get `AttributeError: 'list' object has no attribute 'get'`. Can someone enlighten me about the usage of lines, 21, 22? When I work, with json requests I use the `flask.request.get_json()` as suggested from flask 0.10.

### DIFF
--- a/main/util.py
+++ b/main/util.py
@@ -18,7 +18,7 @@ import config
 ###############################################################################
 def param(name, cast=None):
   value = None
-  if flask.request.json:
+  if flask.request.json and flask.request.method == 'GET':
     return flask.request.json.get(name, None)
 
   if value is None:


gmist:
Hmm... for POST requests you need to set the request content_type to ```application/json```

lipis:
Is that still valid..? I'm thinking of merging the #454..

topless:
I forgot to reply on this one, the `application/json` headers are there as they are supposed to. The problem occurs when I am posting a json array object like `[1, 2, 3]` and try to store it as json property.
If I do like {data: [1, 2, 3]} in can properly iterates on the params. Do I make any sense? or I should try to write it in English. This commit is just skipping the util param when its a POST, its not fixing anything.

topless:
Go ahead and merge 454, much cleaner approach.

PR #470, opened 2016-01-31
Improved api.auth.signing username/email handling

mdxs:
Changed the `username.find('@') > 0` to a `'@' in username` construct, which according to http://stackoverflow.com/a/18437825 is more readable (and less important: faster).

Also, it looked to me that if somehow I have a username with an '@' inside, and provide that as credentials to the `api.auth.signing` end-point, the routine would not find me... thinking that because of the `@` it must be an email address; which if different from my username w/`@` wouldn't be found; and then assuming that there is nothing more to check. I fixed this by allowing the routine to check it as an email, and if it finds it... okay; but if it didn't find a match, it will still check it as a username (and find me).

A further improvement could of course be to actually consider the given arguments and handle these directly, rather then funnelling both username and email through this lookup/check.

### DIFF
--- a/main/api/v1/auth.py
+++ b/main/api/v1/auth.py
@@ -23,15 +23,14 @@ class AuthAPI(restful.Resource):
       'email': wf.Str(missing=None),
       'password': wf.Str(missing=None),
     })
-    username = args['username'] or args['email']
+    handler = args['username'] or args['email']
     password = args['password']
-    if not username or not password:
+    if not handler or not password:
       return flask.abort(400)
 
-    if username.find('@') > 0:
-      user_db = model.User.get_by('email', username.lower())
-    else:
-      user_db = model.User.get_by('username', username.lower())
+    user_db = model.User.get_by(
+      'email' if '@' in handler else 'username', handler.lower()
+    )
 
     if user_db and user_db.password_hash == util.password_hash(user_db, password):
       auth.signin_user_db(user_db)


lipis:
:+1: 

To avoid confusion though.. maybe we should change the variable name to `handler` or something.. as the arguments in the URL are _correct_

```python
handler = args['username'] or args['email']
```

gmist:
``` python
user_db = model.User.get_by('email', username.lower()) if '@' in username\
model.User.get_by('username', username.lower())
```

lipis:
```python
user_db = model.User.get_by('email' if '@' in username else 'username', username.lower())
```

 :D

gmist:
to short ))

lipis:
and shorter if renamed to `handler` ^^

```python
user_db = model.User.get_by('email' if '@' in handler else 'username', handler.lower())
```

gmist:
``` python
user_db = model.User.get_by(['email', 'username']['@' in handler], handler.lower())
```

lipis:
joz3pku

topless:
FU @gmist

gmist:
Fix pep8 warning (#501 line too long)
``` python
x = lambda x: [['username', 'email']['@' in x, x.lower()]
user_db = model.User.get_by(*(x(handler)))
```

main/api/v1/auth.py
```diff
@@ -28,9 +28,10 @@ def post(self):
     if not username or not password:
       return flask.abort(400)
 
-    if username.find('@') > 0:
+    user_db = None
+    if '@' in username:
       user_db = model.User.get_by('email', username.lower())
-    else:
+    if not user_db:
       user_db = model.User.get_by('username', username.lower())
```
lipis: so I would go with:
```python
user_db = model.User.get_by(['username', 'email']['@' in handler], handler.lower())
```

lipis: or if won't fit..

```
user_db = model.User.get_by(
  ['username', 'email']['@' in handler], handler.lower(),
)
```

gmist: I think that it will be more efficient (because it does not allocate memory for temporary array):
``` python
user_db = model.User.get_by('email' if '@' in username else 'username', username.lower())
```

test:
``` python
#!/usr/bin/env python
from timeit import timeit

TEST_DATA='AaAaAaAa@example.com'


def test_1():
  x = ['username', 'email']['@' in TEST_DATA]

def test_2():
  x = 'email' if '@' in TEST_DATA else 'username'


print timeit(test_1, number=1000000)
print timeit(test_2, number=1000000)
```
result:
``` shell
0.251525163651
0.14129281044
```

lipis: lets do that then.. short is nice.. but it's also more readable :)

lipis:
@mdxs Are you around?

topless:
https://m.popkey.co/aa0260/kMR67.gif

mdxs:
Of course, just not online ;-)

mdxs:
That should do the trick; I can do the squashing if so desired.

mdxs:
Ouch, I now realize that the downside of this code reduction is that we're back to the situation that when the given username contains an `@`, it will only be tried as an email address... which sucks if the user provided a valid username with an '@' that doesn't equal that user's email address (perhaps he doesn't even have an email address).

Note the preference for `args[úsername']` when handler is set; and the preference to consider it as an `'email'` in the `get_by()` call if there is an '@' in the handler; thus in the scenario of me having a username with an `@` and no such email address, I will be told to have failed providing proper credentials... even is my username is just perfect.

So now we're back one step... sorry to rush into making suggested changes, I should have been more stubborn.

mdxs:
This approach takes the actual arguments provided and uses that for the database lookup.

It checks that either a `username` or `email` is provided with a `password` (returns 400 if not).

There is at most one db lookup executed (my first approach could execute two): either on `username` (if given) or on `email` (if no `username` is given, but only the `email`). As we previously checked that either `username` or `email` is provided, the `user_db` is known to be either None or set to something when we check the `password`.

If both `username` and `email` are provided, it prefers the db lookup using `username` (if not found, the `email` is NOT considered... so: you better have the `username` right if you provide both).

lipis:
I have to admit that I prefer (and most likely will go with) the:

```python
user_db = model.User.get_by('email' if '@' in handler else 'username', handler.lower())
```

Also from gae-init's point of view.. there is no way to get a username with an `@` :)
https://github.com/gae-init/gae-init/blob/master/main/util.py#L166-L167

mdxs:
Ah, OK... wasn't aware; thought with all those OAuth providers that we could get such exotic usernames.

lipis:
non of the OAuth that we are supporting returning any fancy stuff in their username.. but as already mentioned.. we are supporting only the `.` from the special characters.. the rest is up to the user :)

lipis:
to be more accurate during the automatic one this is happening:


https://github.com/gae-init/gae-init/blob/master/main/auth/auth.py#L366

main/api/v1/auth.py
```diff
@@ -23,15 +23,12 @@ def post(self):
...
     password = args['password']
-    if not username or not password:
+    if not handler or not password:
       return flask.abort(400)
 
-    if username.find('@') > 0:
-      user_db = model.User.get_by('email', username.lower())
-    else:
-      user_db = model.User.get_by('username', username.lower())
+    user_db = model.User.get_by('email' if '@' in handler else 'username', handler.lower())
```
lipis: It's short.. but it's not that short :/

```python
user_db = model.User.get_by(
  'email' if '@' in handler else 'username', handler.lower(),
)
```

PR #473, opened 2016-02-03
Stats

lipis:
Work in progress..

- [x] Stats related model/structure
- [ ] Graphs
- [ ] Stats API
- [ ] Cron jobs
- [ ] Endpoints


mdxs:
+1 for the `isocalendar()` approach.

Something else: I need to check why/how, but the graphs show 2x 2015-W52; probably due to part of that week in 2015 and another part in 2016; should be solvable using ISO week definitions and such.

lipis:
@mdxs I deleted the entries.. but I just realised that:

screen shot 2016-02-05 at 01 46 41

Which kind of sucks!

mdxs:
There is a big performance gain to be made (about 20x faster), with a much simpler and easier to follow `datelimits()` in main/util.py:

```
def date_limits(ts, duration='day'):
  if duration == 'day':
    start = datetime(ts.year, ts.month, ts.day)
    finish = start + timedelta(days=1)
  elif duration == 'week':
    delta = ts.isoweekday() - 1
    start = datetime(ts.year, ts.month, ts.day) - timedelta(days=delta)
    finish = start + timedelta(days=7)
  elif duration == 'month':
    start = datetime(ts.year, ts.month, 1)
    finish = datetime(start.year, start.month + 1, 1)
  elif duration == 'year':
    start = datetime(ts.year, 1, 1)
    finish = datetime(start.year + 1, 1, 1)
  return (start, finish)
```

and then the `import calendar` can be dropped from main/util.py as well

mdxs:
And this `date_code()` is about 3x faster than the current one in main/util.py:

```
def date_code(ts, duration='day', country_code=''):
  if duration == 'day':
    code = '%d-%02d-%02d' % (ts.year, ts.month, ts.day)
  if duration == 'week':
    year, week, day = ts.isocalendar()
    code = '%d-%02d' % (year, week)
  elif duration == 'month':
    code = '%d-%02d' % (ts.year, ts.month)
  elif duration == 'year':
    code = '%d' % (ts.year)
  if country_code:
    return '%s-%s-%s' % (country_code, duration, code)
  return '%s-%s' % (duration, code)
```

mdxs:
hmm, finding some bugs...

mdxs:
Testing with this, it seems that the tasks that start tasks just makes the process very demanding... perhaps I don't fully understand, but I guess the idea is to try and run another task of a higher granularity once one task is completed. Somehow that doesn't quite work out properly... or perhaps I'm just using it wrong, but if I trigger calc/2015 it queues all days in 2015 (fine), but on execution of each of those days (365), a week is queued for that same day (365), which will queue 365 months, which finally queue 365 years... so 365+365+365+365 = 1460 tasks?

lipis:
@mdxs Thanks for the util suggestions..  I updated them..

lipis:
As for the `admin_stats_calc` it was done quick and dirty just to make it work for the current implementation..

*But* in real life you will have to run this once for the whole year or for your past data.. (or recalculating a new property) and once per day it will calculate for yesterday (and eventually for that week, month, year).. so overoptimizing this would not make much sense..

main/util.py
```diff
@@ -217,3 +219,41 @@ def parse_tags(tags, separator=None):
...
+def date_limits(timestamp, duration='day'):
+  if duration == 'week':
+    delta = timedelta(days=timestamp.isoweekday() - 1)
+    start = datetime(timestamp.year, timestamp.month, timestamp.day) - delta
+    finish = start + timedelta(days=7)
+  elif duration == 'month':
+    start = datetime(timestamp.year, timestamp.month, 1)
+    year = start.year + (1 if start.month == 12 else 0)
+    month = (timestamp.month + 12) % 12 + 1
+    finish = datetime(year, month, 1)
```
lipis: @mdxs Fixed now :)

main/auth/auth.py
```diff
@@ -132,6 +132,24 @@ def decorated_function(*args, **kwargs):
...
+      return f(*args, **kwargs)
+    if not is_logged_in() and flask.request.path.startswith('/api/'):
+      return flask.abort(401)
+    if not is_logged_in():
+      return flask.redirect(flask.url_for('signin', next=flask.request.url))
+    return flask.abort(403)
+
+  return decorated_function
+
+
```
lipis: Any better way to have the `cron_required` make the use of the `admin_required`?

mdxs:
I've improved efficiency a lot by scheduling just enough tasks (in the right order), by adjusting `admin_stats_calc()` as follows:

```
@app.route('/admin/stats/calc/<int:year>-<int:month>-<int:day>/')
@app.route('/admin/stats/calc/<int:year>-<int:month>/')
@app.route('/admin/stats/calc/<int:year>/')
@app.route('/admin/stats/calc/')
@auth.admin_required
def admin_stats_calc(year=0, month=0, day=0):
  duration = 'day'
  utc_now = datetime.utcnow()
  if not year:
    year = utc_now.year
    month = utc_now.month
    day = utc_now.day
  if not day:
    duration = 'month'
    day = 1
  if not month:
    duration = 'year'
    month = 1

  ts = datetime(year, month, day)
  start, finish = util.date_limits(ts, duration)
  finish = min(finish, utc_now + timedelta(days=1))

  dct = dict()
  while start < finish:
    for prio, span in enumerate(['day', 'week', 'month', 'year']):
      code = util.date_code(start, duration=span)
      dct.setdefault((prio, code), (prio, start, span))
    start += timedelta(days=1)
  
  tasks = sorted(dct.values())
  for prio, start, span in tasks:
    task.task_calculate_stats(start, span)

  msg = '%d task(s) for the %s' % (len(tasks), util.date_code(ts, duration))
  if util.param('redirect', bool):
    flask.flash('Started %s' % (msg), category='success')
    return flask.redirect(flask.url_for('admin_stats'))
  return msg
```

That will schedule 431 tasks when you request year-2015: 365 + 53 + 12 + 1 in the order of days, weeks, months, and years. Of course, the actual tasks (and scheduling) needs some adjustment in `task.py`:

````
# Stats Related
def task_calculate_stats(ts, duration='day'):
  if ts > datetime(2012, 1, 1) and ts < datetime.utcnow():
    deferred.defer(calculate_stats, ts, duration)


def calculate_stats(ts, duration='day'):
  if duration not in ['day', 'week', 'month', 'year']:
    return

  start, finish = util.date_limits(ts, duration)
  code = util.date_code(ts, duration)
  utc_now = datetime.utcnow()

  stats_db = model.Stats.get_or_insert(
    code,
    parent=config.CONFIG_DB.key,
    code=code,
    timestamp=start,
    duration=duration,
    status='syncing',
  )
  stats_db.status = 'syncing'
  stats_db.put()

  if duration == 'day':
    user_qry = model.User.query()
    user_qry = user_qry.filter(model.User.created >= start)
    user_qry = user_qry.filter(model.User.created < finish)
    #user_dbs = user_qry.fetch(keys_only=True)
    #stats_db.users = len(user_dbs)
    stats_db.users = user_qry.count()
  else:
    stats_qry = model.Stats.query()
    stats_qry = stats_qry.filter(model.Stats.timestamp >= start)
    stats_qry = stats_qry.filter(model.Stats.timestamp < finish)
    if duration == 'year':
      stats_qry = stats_qry.filter(model.Stats.duration == 'month')
    else:
      stats_qry = stats_qry.filter(model.Stats.duration == 'day')
    stats_dbs = stats_qry.fetch()
    
    stats_db.users = 0
    for s_db in stats_dbs:
      stats_db.users += s_db.users

  stats_db.status = 'synced' if finish < utc_now else 'dirty'
  stats_db.put()
```

lipis:
I forgot to to add this line https://github.com/gae-init/gae-init/blob/stats/main/auth/auth.py#L402 from the very beginning.. which explains why the consecutive tasks here: https://github.com/gae-init/gae-init/blob/stats/main/task.py#L210-L215

mdxs:
Calculating statistics whenever a user registers seems a bit heavy... perhaps do those recent ones in memcache or so and add it to the graph results if really needed that fresh?

lipis:
In reality is not that heavy.. if we forget about the first time or old data.. then on every new user.. you will get only the today's users total and then for that week, month, year.. which is not huge or terrible or expensive.. :) Unless of course you have hundreds/thousands users per day... where you'll just disable it and have a cron job on whatever you like..

mdxs:
Okay, fine with me. I was thinking about using the same principle for some other things that I would like to aggregate statistics on; and that would be much more records than those for new users. But I see how I can support that.

main/auth/auth.py
```diff
@@ -132,6 +132,24 @@ def decorated_function(*args, **kwargs):
   return decorated_function
 
 
+def cron_required(f):
+  decorator_order_guard(f, 'auth.cron_required')
+
+  @functools.wraps(f)
+  def decorated_function(*args, **kwargs):
+    if config.PRODUCTION and 'X-Appengine-Cron' in flask.request.headers:
```
gmist: It is safe?

lipis: According to the [docs](https://cloud.google.com/appengine/docs/python/config/cron?hl=en#Python_app_yaml_Securing_URLs_for_cron) yes.. 

>The X-Appengine-Cron header is set internally by Google App Engine. If your request handler finds this header it can trust that the request is a cron request. If the header is present in an external user request to your app, it is stripped, except for requests from logged in administrators of the application, who are allowed to set the header for testing purposes.

> Google App Engine issues Cron requests from the IP address 0.1.0.1.

lipis:
The idea is to continue adding more properties to the Stats model.. and calculating it in the same task.. and of course if it gets complicated and stuff.. act accordignly.. but here I want to have a dead simple/straight forward way without too much complexity.. :) at least as the first step..

main/auth/auth.py
gmist: I missed this point. Thank you!

lipis: It's actually pretty awesome that they support that.. :)

main/auth/auth.py
mdxs: Feels like there should be two endpoints; one for admins to start calculations manually (using `admin_required`) and another one that uses (a simplified) `cron_required` for cron only (perhaps in non-PRODUCTION allow admins for debugging/testing). In that setup, the end-points just call upon a worker method: basically the current `admin_stats_calc` method, but perhaps moved (in its generic form) to `task.py` under Stats Related.

The cron end-point may not take parameters, as `when=yesterday` would probably be the only arguments anyway. It would call the generic method based on values from `now - 1 day`.

lipis: Some could argue that you could have more cron jobs that a repeated every week or month or year.. depending on the use case and combining the cron/admin rights is not a bad idea as you can simplify the testing from the same URL, plus you can add those admin endpoints somewhere in your admin console for manual sync..

mdxs:
Understand the adding more properties to Stats model; that is my plan as well. In fact, why is there a `code` property that has the same value as Key Name?

lipis:
What [code](https://github.com/gae-init/gae-init/commit/532e2a83b4d124cbaf7d9eaf5b068c9675fe74d4)?

main/auth/auth.py
mdxs: Perhaps a small difference, using:

```
def cron_required(f):
  decorator_order_guard(f, 'auth.cron_required')

  @functools.wraps(f)
  def decorated_function(*args, **kwargs):
    if config.PRODUCTION:
      if 'X-Appengine-Cron' in flask.request.headers:
        # TODO: check IP == 0.1.0.1
        return f(*args, **kwargs)
    else:
      if is_logged_in() and current_user_db().admin:
        return f(*args, **kwargs)
      if not is_logged_in():
        return flask.redirect(flask.url_for('signin', next=flask.request.url))
    return flask.abort(403)

  return decorated_function
```

To only allow admin access in non-PRODUCTION environment; also note that I removed the `/api/` check, as in my world that shouldn't be a cron end-point.

lipis: My question was more like if I could use somehow the `admin_required(f)` inside the `cron_required` from the decorator's point of view.. never dived into the decorators world that much :)

lipis: Plus if the App engine guarantees that they will strip down the `X-Appengine-Cron` why checking for IP..?

mdxs: Understand, neither did I; however I tested some more and this works locally on the devserver:

```
def cron_required(f):
  decorator_order_guard(f, 'auth.cron_required')

  @functools.wraps(f)
  def decorated_function(*args, **kwargs):
    if 'X-Appengine-Cron' in flask.request.headers:
      ip = flask.request.remote_addr
      if ip == '0.1.0.1':
        return f(*args, **kwargs)
    return flask.abort(403)

  return decorated_function
```

When associated with an endpoint, the devserver shows them as Cron Jobs and you can issue a Run Now; and then the unique header/ip combo is like you would see on production.

lipis: but but but.. this one you won't be able to invoke manually using a direct URL.. even if you're an admin

mdxs: Correct, as said, I like to split the responsibility in this case. Using the following `control/cron.py`:

```
# coding: utf-8

from datetime import datetime
from datetime import timedelta

import auth
import task

from main import app


@app.route('/cron/stats/calc/<string:when>/')
@auth.cron_required
def cron_stats_calc(when=''):
  if when == 'daily':
    ts = datetime.utcnow() + timedelta(-1)
    tasks, msg = task.schedule_stats_calc(ts.year, ts.month, ts.day)
    return msg, 200
  return '', 403
```

Where the `task.schedule_stats_calc` is the shared common worker method with the admin controlled manual calculation endpoints.

lipis: and losing the whole magicness of gae-init.. where things are happening securely and out of the box :) I'm not convinced that supporting the cron job that is hard to test it in production... is better than just combining the headers and `admin_required`.. super simple.. and even if it's not optimised or have duplicate code..

the benefits of being able to test any cron job without a cron job or special deployments is big :)

mdxs: Sure, no problem; just sharing my findings. You may want to see what the Cron Jobs page on the devserver reports when you use Run Now for such jobs. Perhaps you never will, but newbies may. Note I'm just lazy now... as I made my own stuff, I cannot easily test this with gae-pure ;-)

PR #487, opened 2016-02-24
Fixed missing marshmallow dependency

topless:


gmist:
It's already in [webargs](https://github.com/sloria/webargs/blob/dev/setup.py#L6).

topless:
Indeed, but it required a `gulp clean:venv` after my merge to grab it.

PR #490, opened 2016-03-09
Implemented AuthSignupApi.

topless:


lipis:
Interesting.. but but but what about ze bots?

giphy

lipis:
I'll pass on this one for now..

Issue #493, opened 2016-03-18
Prevent gulp-cssnano from changing z-index by default


dhstack:
I spent way to many hours trying to figure out why the local hosted version of my site worked fine and the deployed version was not when it came to displaying certain elements. It turns out gulp-cssnano changes z-index values ***by default***. I suggest we change this behavior, as [this issue](https://github.com/ben-eb/gulp-cssnano/issues/8) suggests, and pass `{zindex: false}` to cssnano until version 4 is released that changes this default behavior.

mdxs:
Could you propose what to change where in a Pull Request?

dhstack:
Yes, working on that now

mdxs:
Thnx, looking forward; though I didn't experience the issue myself.

dhstack:
It may only happen with large values for the z-index. I had some of mine over 100

topless:
I didn't experience that either and I as far as I am aware bootstrap classifies `z-index` of its elements like 1000, 1010, 1020, but I see also its part of `cssnano` optimizations, nice catch, what os were you developing on if you don't mind?

dhstack:
Windows 10

dhstack:
Since this is now merged, I'm closing this issue.

PR #495, opened 2016-03-18
Fix for zindex values being changed by gulp-cssnano by default

dhstack:
Fixes issue #493 to prevent z-index values in css files to automatically be changed by gulp-cssnano


lipis:
Since it will be fixed later on.. I wouldn't merge it normally.. mostly because it will be forgotten and obsolete with the version 4... So you will be responsible for removing this line when that happens.

dhstack:
Noted.

PR #417, opened 2015-09-27
Updating Bitbucket to OAuth2

lipis:
https://confluence.atlassian.com/bitbucket/oauth-on-bitbucket-cloud-238027431.html

But I'm keep getting `unauthorized_client` as a response. Any ideas?

PR #496, opened 2016-03-19
Updated Bitbucket to OAuth2

mdxs:
Updated Bitbucket to OAuth2 interface, see https://confluence.atlassian.com/bitbucket/oauth-on-bitbucket-cloud-238027431.html (and previous attempt in #417).

On the OAuth tab of the Bitbucket settings (of your profile there), you can configure "OAuth consumers". Provide a "Callback URL" to use the OAuth2 interface (this PR includes the link on the Auth Config part for Bitbucket); and set the Email and Read permissions for Account.

Note that there is **no** `request_token_params={'scope': 'account'}` being used; as Bitbucket noted:

> Scopes are defined on the client/consumer instance. Bitbucket Cloud does not currently support the use of the optional scope parameter on the individual grant requests.

_Background for this PR:_ I noticed that using the OAuth1 interface, I needed to set much more permissions (also Read on Team membership and Read on Repositories) for no good reason other than to make it work (such that gae-init allows a new user to register using Bitbucket).

### DIFF
--- a/main/auth/bitbucket.py
+++ b/main/auth/bitbucket.py
@@ -10,12 +10,12 @@ import util
 from main import app
 
 bitbucket_config = dict(
-  access_token_url='https://bitbucket.org/api/1.0/oauth/access_token',
-  authorize_url='https://bitbucket.org/api/1.0/oauth/authenticate',
-  base_url='https://api.bitbucket.org/1.0/',
+  access_token_method='POST',
+  access_token_url='https://bitbucket.org/site/oauth2/access_token',
+  authorize_url='https://bitbucket.org/site/oauth2/authorize',
+  base_url='https://api.bitbucket.org/2.0/',
   consumer_key=config.CONFIG_DB.bitbucket_key,
   consumer_secret=config.CONFIG_DB.bitbucket_secret,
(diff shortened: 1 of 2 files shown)


main/auth/bitbucket.py
```diff
@@ -10,12 +10,13 @@
...
-  authorize_url='https://bitbucket.org/api/1.0/oauth/authenticate',
-  base_url='https://api.bitbucket.org/1.0/',
+  access_token_method='POST',
+  access_token_url='https://bitbucket.org/site/oauth2/access_token',
+  authorize_url='https://bitbucket.org/site/oauth2/authorize',
+  base_url='https://api.bitbucket.org/2.0/',
   consumer_key=config.CONFIG_DB.bitbucket_key,
   consumer_secret=config.CONFIG_DB.bitbucket_secret,
-  request_token_url='https://bitbucket.org/api/1.0/oauth/request_token',
+  #request_token_params={'scope': 'account'},
```
mdxs: I'll remove this commented out line; as Bitbucket noted:

> Scopes are defined on the client/consumer instance. Bitbucket Cloud does not currently support the use of the optional scope parameter on the individual grant requests.

PR #512, opened 2016-05-16
Fixed callback URL for Facebook

mdxs:
Fixed callback URL for Facebook, to be used when configuring your App on the Facebook for Developers site, under the "Valid OAuth redirect URIs" of the "Client OAuth Settings" section.


lipis:
I'm guessing you verified it.. :)

mdxs:
Yes, I did.

PR #515, opened 2016-05-29
Flask 0.11

lipis:
I'm still not sure about the `import flask_wtf as wtf`. I'm leaning towards the `import flask_wtf` instead..

### DIFF
--- a/main/__init__.py
+++ b/main/__init__.py
@@ -10,4 +10,4 @@ https://gae-init.appspot.com
 
 """
 
-__version__ = '4.1.3'
+__version__ = '5.0.0'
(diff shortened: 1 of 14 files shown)


gmist:
I would prefer wtf...

lipis:
In general I don't like the `...as something` because you always have to remember when importing it.. I get the backwards compatibility issues that will occur but if you have new files the `from flask.ext import wtf` is still going to work..

Also if we would start this project today.. most likely we would use the `import flask_restful`, `import flask_wtf`, etc.. without using the shorthand versions :)

Life decisions :D

giphy

lipis:
@gmist you know that I respect your opinion.. but I'll go ahead with the version without `as` on this one :)

PR #516, opened 2016-05-30
Added bash scripts

lipis:
Kudos to @mdxs and his idea.. 

### Using Git in  [Cloud Shell](https://cloud.google.com/shell/)

```bash
git clone -b console https://github.com/gae-init/gae-init.git
./gae-init/bin/requirements_cloud_shell.sh
```

####  [Cloud Shell](https://cloud.google.com/shell/)
```
curl https://raw.githubusercontent.com/gae-init/gae-init/console/bin/requirements_cloud_shell.sh | bash
```

#### Linux
```
curl https://raw.githubusercontent.com/gae-init/gae-init/console/bin/requirements_linux.sh | bash
```

##### Homebrew

```
curl https://raw.githubusercontent.com/gae-init/gae-init/console/bin/requirements_osx_brew.sh | bash
```

##### MacPorts

```
curl https://raw.githubusercontent.com/gae-init/gae-init/console/bin/requirements_osx_port.sh | bash
```


topless:
Splendid

bin/console_requirements.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Node.js
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+  sudo apt-get install nodejs
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+  brew install node
+fi
+
+# Gulp.js
+sudo npm install -g gulp
```
gmist: homebrewed OS X is not required ```sudo```

lipis: good catch.. thanks!

lipis:
Should I add something like that to the beginning of it? Or Homebrew should be just mentioned.. as it's not a real requirement..

```bash
if [[ "$OSTYPE" == "darwin"* ]]; then
  which -s brew
  if [[ $? != 0 ]]; then
    /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
  else
    brew update
  fi
fi
```

lipis:
Or maybe just informing brew is not present?

```
echo "Homebrew is required on OS X in order to continue with the script, install it via http://brew.sh/ or install all the requirements manually: http://docs.gae-init.appspot.com/requirement/"
```

bin/requirements_cloud_shell.sh
```diff
@@ -0,0 +1,16 @@
...
+# Init
+mkdir -p ~/temp/gae-init-temp
+cd ~/temp/gae-init-temp
+
+# Gulp.js
+sudo npm install -g gulp
+
+# Python related
+curl -O https://bootstrap.pypa.io/get-pip.py
+sudo python get-pip.py
```
lipis: maybe even `pip` is already preinstalled..

gmist: yep

gmist:
[MacPorts](https://www.macports.org) also can be used... or [Nix](http://ariya.ofilabs.com/2016/05/nix-as-os-x-package-manager.html)...

lipis:
I might pass on Nix until more people will ask for it in the future.. but MacPorts should be added..

lipis:
@gmist @mdxs @topless am I missing something?

gmist:
looks good

PR #525, opened 2016-06-08
AngularJS

lipis:
### gae-init-angularjs

This will not be merged but eventually will be most likely under it's own project called **gae-init-angularjs** which will be for Angular 1. The main goal is to have **Angular 2** and **TypeScript** out of the box.. this is just a stepping point and for anyone wants to use Angular in their app.
- [x] Bower updates
- [x] Gulp updates
- [ ] General Setup
- [ ] Admin Console
- [x] User List
- [ ] User View
- [x] Config View
- [ ] Config Updated
- [ ] Config Auth View
- [ ] Config Auth Update
- [ ] User Profile
- [ ] Feedback
- [x] `magic.py`
- [ ] Templates
- [ ] Routing
- [ ] Header
- [ ] etc..

---
### Testing things

If you are familiar with Git then you most likely know already what to do in order to get this `angularjs` branch. If not here is the simplest way:

```
git clone https://github.com/gae-init/gae-init.git
cd gae-init
git checkout angularjs
npm install
gulp
```

If no errors occurred visit [localhost:8080/angular/](http://localhost:8080/angular/). You need to login as admin in order to see the actual data.
### Magic integration

Once everything is there and working locally, from the root folder you can use `magic.py` to generate the scafolding for more models.

Here is an example for our [Issue Tracker](http://magic.gae-init.appspot.com/project/5798568036139008/):

``` bash
./magic.py -p 5798568036139008
gulp
```

screen shot 2016-06-09 at 11 08 33

You can create your [own projects](http://magic.gae-init.appspot.com/project/) and test them using the `magic.py` script like above.

---

P.S. All the commits most likely will be squashed in the end..

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,10 @@
 {
   "name": "gae-init",
   "dependencies": {
+    "angular": "~1",
+    "angular-moment": "~1",
+    "angular-resource": "~1",
+    "angular-route": "~1",
     "bootstrap": "~3",
     "font-awesome": "~4",
     "jquery": "~3",
--- a/gulp/config.coffee
+++ b/gulp/config.coffee
@@ -4,6 +4,10 @@ config =
   ext: [
     "#{paths.static.ext}/jquery/dist/jquery.js"
     "#{paths.static.ext}/moment/moment.js"
+    "#{paths.static.ext}/angular/angular.js"
+    "#{paths.static.ext}/angular-moment/angular-moment.js"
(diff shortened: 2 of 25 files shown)


phivo:
I'm glad you have started this, and hope this can be developed further. Nowadays, the trend is to have a client-side front-end, and having a current project that does not reflect this sounds too old-school. And it only makes sense as the back-end api is already there.

Issue #526, opened 2016-06-08
ImportError: No module named msvcrt


dhstack:
Because [Click](http://click.pocoo.org/5/) was added to Flask 0.11 for the [New Command Line Interface](http://flask.pocoo.org/docs/0.11/cli/), and Click imports the Windows only module [msvcrt](https://docs.python.org/2/library/msvcrt.html), and the cloud devserver doesn't contain this module, an import error is thrown.

![image](https://cloud.githubusercontent.com/assets/2013388/15913823/3b48f59e-2d91-11e6-8442-c79248d82e3c.png)

I'll create a PR with the workaround I adapted from [this Stackoverflow question](http://stackoverflow.com/a/25915535/2943925).

dhstack:
These are all the file locations that Click tries to import or use msvcrt

![image](https://cloud.githubusercontent.com/assets/2013388/15914088/e95477b6-2d92-11e6-9f2c-b0f192888bee.png)

pvlele:
appengine_config.py


import os
import sys

if os.environ.get('SERVER_SOFTWARE', '').startswith('Google App Engine'):
  sys.path.insert(0, 'lib.zip')
else:
  if os.name == 'nt':
    os.name = None
    sys.platform = ''

  import re
 # from google.appengine.tools.devappserver2.python import stubs
  from google.appengine.tools.devappserver2.python.runtime import stubs

  re_ = stubs.FakeFile._skip_files.pattern.replace('|^lib/.*', '')
  re_ = re.compile(re_)
  stubs.FakeFile._skip_files = re_
  sys.path.insert(0, 'lib')
sys.path.insert(0, 'libx')


def webapp_add_wsgi_middleware(app):
  from google.appengine.ext.appstats import recording
  app = recording.appstats_wsgi_middleware(app)
  return app
this solved my problem

mdxs:
@pvlele Could you edit your comment and use code quotes to preserve formatting?

PR #527, opened 2016-06-08
Fix for msvcrt import error (issue #526)

dhstack:
Fixes issue #526

### DIFF
--- a/main/appengine_config.py
+++ b/main/appengine_config.py
@@ -6,6 +6,10 @@ import sys
 if os.environ.get('SERVER_SOFTWARE', '').startswith('Google App Engine'):
   sys.path.insert(0, 'lib.zip')
 else:
+  if os.name == 'nt':
+    os.name = None
+    sys.platform = ''
+
   import re
   from google.appengine.tools.devappserver2.python import stubs
 


main/appengine_config.py
```diff
@@ -6,6 +6,10 @@
 if os.environ.get('SERVER_SOFTWARE', '').startswith('Google App Engine'):
   sys.path.insert(0, 'lib.zip')
 else:
+  if os.name == 'nt':
+    os.name = None
+    sys.platform = ''
```
lipis: In the SO question there is nothing about `sys.platform`.. why is that?

dhstack: Because in that question it was for Django and only os.name was used to check if the os was windows. In this case, Click uses both sys.platform and os.name to check if the os is Windows. Using only os.name doesn't work.

See
https://github.com/pallets/click/search?utf8=%E2%9C%93&q=sys.platform&type=Code
and
https://github.com/pallets/click/search?utf8=%E2%9C%93&q=nt&type=Code

lipis: Verified on mine.. it works like this.. thanks!

Maybe we should file an issue on click for that matter.. WDYT?

dhstack: No problem!

Yeah, we probably should file an issue with Click as it is a part of Flask now, and I'm sure we're not the only ones who are going to run into this issue with App Engine...in which case, it would probably be good to file one with Flask too.

lipis: @dhstack Will you do that..? and post here the issue numbers so we could vote/subscribe..

dhstack: Done.

Flask issue:
https://github.com/pallets/flask/issues/1901

Click issue:
https://github.com/pallets/click/issues/594

yassirSNS:
I am facing same issue, it's annoying . I'm new to Flask on App Engine.

First dev_appserver.py app.yaml command was not running, it opens dev_appserver.py source code in editor. -> I changed associate file for .py to default python.exe, now working, but to open source code of .py file, now I have to right click and open with any editor (annoying).

Secondly it gives error dev_appserver: too few argumunets -> added /* in registry setting (No idea what it does). working now.

Now this Click thing. -.> installed Click 5.1 using pip, still issue is there.
raise ImportError('No module named %s' % fullname)
ImportError: No module named msvcrt
This fixed by your code.

Now another issue-
from flask import Flask
ImportError: No module named flask

thogh flask is installed lib folder as well as in \env\Lib\site-packages. 


Please help.

PR #537, opened 2016-06-18
Fixed test pagination

mdxs:
There was a broken backlink in the Pagination part of the Test page, replaced it with an example of providing a custom caption previous page link.


lipis:
Same here.. #540

PR #538, opened 2016-06-18
Fixed test page test section hover links

mdxs:
The on-hover "Just this test" links next to the title of test sections on the test page didn't work (anymore) and caused 500 errors; fixed by changing them to "Back to top" on-hover links.


lipis:
#540

lipis:
maybe we could change the copy or something.. but that was the idea :)

mdxs:
I see; and it works now, so I'll abandon this PR.

PR #541, opened 2016-06-24
Moved coffeescript, less and uglify-js to npm devDependencies.

topless:


package.json
```diff
@@ -10,12 +10,10 @@
     "install": "gulp init"
   },
   "dependencies": {
-    "coffee-script": "~1",
-    "less": "~2",
-    "uglify-js": "~2"
   },
```
lipis: then remove the the whole `dependencies` section as well

PR #542, opened 2016-06-24
Installing some front-end libs with npm

lipis:


gmist:
Hmm... what for?

lipis:
Mostly experimental.. because for example some libraries like Angular 2 don't support Bower (or maybe they do) and in general if we can get away with one would be nice.. it looks like we can't get rid of it though for libraries that require more static data.. :)

Also even though I really like Bower and I agree with [calidion's answer](https://github.com/bower/bower/issues/2298#issuecomment-222927708) more and more people are just using npm and I wanted to check how it could be..

lipis:
After some thinking there is no reason to confuse the front end with the build tools, plus a few other things that are not really solving anything...

Issue #549, opened 2016-08-03
Cannot install on windows


phivo:
Followed all prerequisites, `npm install` was successful, but running `gulp` was not.
It throws an error on module.js:422, saying that cannot find module `lodash.keys`.

Please help. Error code below

```
Error: Cannot find module 'lodash.keys'
  at Function.Module._resolveFilename (module.js:440:15)
  at Function.Module._load (module.js:388:25)
  at Module.require (module.js:468:17)
  at require (internal/module.js:20:19)
  at Object.<anonymous> (C:\Users\xxxxxx\coding\gae-init\node_modules\lodash.template\index.js:14:12)
  at Module._compile (module.js:541:32)
  at Object.Module._extensions..js (module.js:550:10)
  at Module.load (C:\Users\xxxxxx\coding\gae-init\node_modules\coffee-script\lib\coffee-script\register.js:45:36)
  at tryModuleLoad (module.js:417:12)
  at Function.Module._load (module.js:409:3)
  at Module.require (module.js:468:17)
  at require (internal/module.js:20:19)
  at Object.<anonymous> (C:\Users\xxxxxx\coding\gae-init\node_modules\gulp-util\lib\template.js:1:78)
  at Module._compile (module.js:541:32)
  at Object.Module._extensions..js (module.js:550:10)
  at Module.load (C:\Users\xxxxxx\coding\gae-init\node_modules\coffee-script\lib\coffee-script\register.js:45:36)
  at tryModuleLoad (module.js:417:12)
  at Function.Module._load (module.js:409:3)
  at Module.require (module.js:468:17)
  at require (internal/module.js:20:19)
  at Object.<anonymous> (C:\Users\xxxxxx\coding\gae-init\node_modules\gulp-util\index.js:7:13)
  at Module._compile (module.js:541:32)
  at Object.Module._extensions..js (module.js:550:10)
  at Module.load (C:\Users\xxxxxx\coding\gae-init\node_modules\coffee-script\lib\coffee-script\register.js:45:36)
  at tryModuleLoad (module.js:417:12)
  at Function.Module._load (module.js:409:3)
  at Module.require (module.js:468:17)
  at require (internal/module.js:20:19)
  at Object.<anonymous> (C:\Users\xxxxxx\coding\gae-init\node_modules\gulp\index.js:5:13)
  at Module._compile (module.js:541:32)
  at Object.Module._extensions..js (module.js:550:10)
[...]

Issue #543, opened 2016-06-26
"OSError: [Errno 13] Permission denied: '/lib'" when running npm install in OS X


impiaaa:
After following the [Requirements](http://docs.gae-init.appspot.com/requirement/) page manually, I continued to run `npm install`. It gets to `pip install -q -r requirements.txt`, then spits out this error:

```
Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/local/lib/python2.7/site-packages/pip/commands/install.py", line 317, in run
    prefix=options.prefix_path,
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_set.py", line 742, in install
    **kwargs
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_install.py", line 831, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_install.py", line 1032, in move_wheel_files
    isolated=self.isolated,
  File "/usr/local/lib/python2.7/site-packages/pip/wheel.py", line 346, in move_wheel_files
    clobber(source, lib_dir, True)
  File "/usr/local/lib/python2.7/site-packages/pip/wheel.py", line 287, in clobber
    ensure_dir(dest)  # common for the 'include' path
  File "/usr/local/lib/python2.7/site-packages/pip/utils/__init__.py", line 83, in ensure_dir
    os.makedirs(path)
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/lib'
```

[...]

Issue #510, opened 2016-05-14
error when trying to deploy to application


ynotlim:
I'm getting this error when trying to deploy my application.  Any ideas why this error is coming up?  No issues when running it locally.

Thanks in advance!

![capture](https://cloud.githubusercontent.com/assets/3355351/15266128/296cbbe4-1969-11e6-8bdc-1e8fbeb8530c.PNG)

lipis:
is it the latest gcloud?

ynotlim:
yeah, latest gcloud.  I was able to fix it by changing `main/*.yaml `to `main/app.yaml`  may be a Windows issue.

ynotlim:
with this update should build.cofffee now be?

```
cmd: "gcloud preview app deploy main/index.yaml #{options_str}"'
cmd: "gcloud preview app deploy main/app.yaml #{options_str}"'
```

lipis:
but there are also `queue.yaml`, `cron.yaml` (maybe more 😕).. so that's why it was `*.yaml`... we'll have to check the general solution for Windows if that's the case..

lipis:
but it's also the `queue.yaml`

ynotlim:
FYI... I tried it on powershell and cmd, same error

```
c:\Users\limtc\Documents\GitHub\gae-init\main>gcloud preview app deploy *.yaml
ERROR: File [*.yaml] not found.
ERROR: (gcloud.preview.app.deploy) Errors occurred while parsing the App Engine app configuration.

```

lipis:
might file a bug for gcloud :)

dhstack:
I'm having this same issue on Windows 10

dhstack:
@ynotlim proposed fix in build only appears to be updating whichever .yaml file is last on the list.

So,

```
cmd: "gcloud preview app deploy main/index.yaml #{options_str}"'
cmd: "gcloud preview app deploy main/app.yaml #{options_str}"'
```
only updates app.yaml. If you switch it, only index.yaml is updated.

lipis:
Maybe
```bash
gcloud preview app deploy main/app.yaml main/index.yaml #{options_str}
```

akashawasthi:
[This](https://cloud.google.com/appengine/docs/flexible/java/dev-java-only#deploying_your_app) (under a _Java_ doc) advises to deploy any _optional_ configuration files separately. I guess it makes some sense intuitively.
>
[...]

Issue #492, opened 2016-03-15
Running 'gulp' command just open the dev_appserver.py file


alex-deneuvillers:
After running 'gulp' on my terminal (Windows 7) I get the following : 

> [11:42:32] Using gulpfile ~\PycharmProjects\Project\gulpfile.js
> [11:42:32] Starting 'default'...
> [11:42:32] Starting 'run'...
> [11:42:32] Starting 'init'...
> [11:42:32] Starting 'pip'...
> [11:42:32] Finished 'run' after 203 ms
> [11:42:32] Starting 'watch'...
> [11:42:32] Finished 'watch' after 131 ms
> [11:42:32] Starting 'reload'...
> [11:43:20] Finished 'reload' after 47 s
> [11:43:20] Finished 'default' after 48 s
> [11:43:20]  $ python run.py -d
> [11:43:34] Finished 'pip' after 1.02 min
> [11:43:34] Starting 'bower'...
> [11:43:34]  $ node_modules.bin\bower install
> [11:43:47] Finished 'bower' after 14 s
> [11:43:47] Starting 'copy_bower_files'...
> [11:43:51] Finished 'copy_bower_files' after 3.42 s
> [11:43:51] Finished 'init' after 1.3 min
> [11:43:51] Starting 'ext:dev'...
> [11:43:51] Starting 'script:dev'...
> [11:43:52] Starting 'style:dev'...
> [11:44:06] Finished 'style:dev' after 14 s
> [11:44:06] Finished 'script:dev' after 15 s
> [11:44:06] Finished 'ext:dev' after 15 s
> [11:44:06] C:\Some\Path\Project\main\static\dev\script\ext.js reloaded.
> [11:44:06] C:\Some\Path\Project\main\static\dev\script\script.js reloaded.
> [11:44:06] C:\Some\Path\Project\main\static\dev\style\style.css reloaded.
> [11:44:07]  $ python run.py -s

And instead of launching the dev server, it just opens the dev_appserver.py file in my editor.

lipis:
is this still an issue?

alex-deneuvillers:
Yeah, I still have this issue. The watch commands works fine but the devappserver.py just opens so I have to use the Google App Engine Launcher.

lipis:
Could you please install the latest version of the `gcloud` from: https://cloud.google.com/sdk/ and also get a fresh copy of `gae-init` and try it again?

alex-deneuvillers:
[...]

PR #550, opened 2016-08-09
Dockerize gae-init

gmist:
Docker way to preventing issues like #549, #543, #526, #510, #492, #491, #471, #460...


lipis:
Quite interesting.. but on vacations now :) definitely promising..

run.py
```diff
@@ -190,6 +167,17 @@ def _exclude_ext(pkg):
         return True
     return False
 
+  if dir_ in exclude or _exclude_prefix(dir_) or _exclude_ext(dir_):
+    return True
+  return False
```
lipis: or just 

```python
dir_ in exclude or _exclude_prefix(dir_) or _exclude_ext(dir_)
```

run.py
```diff
@@ -257,83 +243,16 @@ def print_out_update(force_show=False):
...
 
 
-###############################################################################
-# Doctor
-###############################################################################
 def internet_on():
   try:
-    urllib2.urlopen(INTERNET_TEST_URL, timeout=2)
+    h = urllib2.urlopen(INTERNET_TEST_URL, timeout=2)
+    h.close()
```
lipis: maybe `handle` to avoid the single character...

package.json
```diff
@@ -31,6 +31,7 @@
     "gulp-start": "~1",
     "gulp-uglify": "~2",
     "gulp-util": "~3",
+    "gulp-watch": "~4",
```
lipis: Do you want to have that as a separate PR to the core?

gmist: It is not necessary for the core - is just a way to workaround [this problem](https://forums.docker.com/t/docker-compose-volumes-gulp-watch-100-cpu-usage/10192).

lipis: There is always something! 💃

irvingpop:
This is super slick and I love it.   I only needed to make a few minor changes to the docker-gae-init container ( https://github.com/irvingpop/docker-gae-init/commit/06d259859c16afb218ead745d89df493de9470c0 ) and I'm happily running gae-init unmodified.

This is an excellent solution for onboarding developers without requiring significant local system modifications.

lipis:
Travis is failing :)

PR #562, opened 2016-09-12
Replacing 'do' with '()'

lipis:
@gmist I hope you are not mad :) But `do` is a bit confusing and not very common I would say..


gmist:
qewt21
))

PR #567, opened 2016-09-25
Make historical starting date inclusive

mdxs:
The historical starting date of 2012-01-01 should be included to avoid not having statistics for that date... Proposing to change (even if you did only party that day, or sleeping; or whatever) as others implementing stats should by default include such a historical/theoretical starting date; which they should probably set (close) to their launch date.

### DIFF
--- a/main/auth/auth.py
+++ b/main/auth/auth.py
@@ -397,6 +397,7 @@ def create_user_db(auth_id, name, username, email='', verified=False, **props):
   )
   user_db.put()
   task.new_user_notification(user_db)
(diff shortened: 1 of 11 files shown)


mdxs:
Note that this belongs in the #473 Stats example.

lipis:
I would prefer that to be a comment instead of PR.. as it's very very small change and usually the whole `example` is squashed :D changed it there..

mdxs:
Okay, will try to remember next time ;-)

lipis:
💃 🍻 📈 📊

PR #568, opened 2016-09-25
Removed extra .bowerrc file.

topless:


lipis:
yeah.. I think it's more or less useless :)

PR #576, opened 2016-10-15
Added whitespace to parameters

mdxs:
Made this line similar to https://github.com/gae-init/gae-init/blob/master/main/static/src/style/signin.less#L22


lipis:
The other line has to be fixed.. :)

http://codeguide.co/#css-syntax

mdxs:
I see: 

> Don't include spaces after commas within rgb(), rgba(), hsl(), hsla(), or rect() values. This helps differentiate multiple color values (comma, no space) from multiple property values (comma with space).

Will open a different PR.

PR #578, opened 2016-10-16
Renamed flask_wtf.Form to .FlaskForm

mdxs:
This is to addresses the following deprecation warning from Flask-WTF as seen in the logs:

> FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.


lipis:
💯

PR #579, opened 2016-10-16
Update webargs to v1.4.0

mdxs:


lipis:
There is always one more thing.. 30 seconds after a new release! :D

mdxs:
Note that after the webargs v1.4.0 release, there was further discussion of handling errors in Flask-RESTful starting with https://github.com/sloria/webargs/issues/122#issuecomment-253413919 ... which I must admit I cannot fully follow right now. So this PR may need some further testing anyway.

Issue #580, opened 2016-10-21
High CPU consumption when running gulp locally


ventice:
**Platform:** Windows
**Steps to reproduce:**
1. Run local server by typing `gulp`
2. Open task manager and find `node.exe` process

**Result:**
node.exe consumes about 100% of CPU
**Expected result:**
node.exe consumes less than 25% of CPU

**Side notes:**
The issue is caused by gulp.watch for livereload. As reported in [Isse#634 of gulp.watch](https://github.com/gulpjs/gulp/issues/634), this can be mitigated by increasing watch interval.

gmist:
You can use ```gulp-watch``` instead ```gulp.watch```:
1. [npm install gulp-watch --save-dev](https://github.com/gae-init/gae-init/pull/550/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R34)
2. [replace ```watch``` by ```gulp.watch```](https://github.com/gae-init/gae-init/pull/550/files#diff-724f8714ebe98bf7b20b1fb62e8ca522R20)

ventice:
I already fixed it by increasing polling interval as recommended in the cited issue. Creating pull request.

lipis:
Hey @ventice I tried the gulp-watch as recommended from @gmist so there's the PR #583 can you check if the CPU still high with it? It should be better now...

ventice:
Ok, will check tomorrow and keep you posted.

ventice:
Took `package.json` and `watch.coffee` from given PR and started gulp. Got a warning and looped watching of `ext`:

[15:22:34] Starting 'ext:dev'...
(node) warning: possible EventEmitter memory leak detected. 11 task_stop listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
  at Gulp.addListener (events.js:239:17)
  at C:\Users\venticello\GoogleApps\opttorg-api\default\node_modules\gulp-sequen
ce\index.js:58:12
  at apply (C:\Users\venticello\GoogleApps\opttorg-api\default\node_modules\gulp-sequence\node_modules\thunks\thunks.js:338:38)
  at tryRun (C:\Users\venticello\GoogleApps\opttorg-api\default\node_modules\gulp-sequence\node_modules\thunks\thunks.js:217:19)
  at runThunk (C:\Users\venticello\GoogleApps\opttorg-api\default\node_modules\gulp-sequence\node_modules\thunks\thunks.js:210:15)
[...]

PR #581, opened 2016-10-21
Increased watch polling interval to save CPU

ventice:
This PR fixes issue [#580](https://github.com/gae-init/gae-init/issues/580). By default interval is 100ms, increased it to 500ms.


ventice:
Closing because of #583

PR #582, opened 2016-10-21
Now after npm install and gulp init the gulp command will execute aut…

topless:
…omatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "install": "gulp init",
-    "postinstall": "gulp"
+    "postinstall": "echo 'Run `gulp` to start or `gulp help` for more.'"
   },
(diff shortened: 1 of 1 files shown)


topless:
As lazy as it gets as you can see..

lipis:
Maybe we could print on how to start there but with the `npm install` I don't want anything to start that doesn't end..

Something like that? WDYT?

```json

"postinstall": "echo 'Run `gulp` to start or `gulp help` for more.'"
```

topless:
That could work too, to display the next step you have to take. I was just showing off about `postinstall`

topless:
It prints the command twice, once as a postinstall command and once the echo of the postinstall command. If you like it we keep it.

screen shot 2016-10-24 at 15 46 09

PR #599, opened 2016-12-03
Added Flask-Talisman (HTTP security headers)

lipis:


lipis:
This most likely is going to end up in the core.. everyone should be more secure these days., and SSL is for free :)

lipis:
I'll merge that one.. if there are any objections.. speak now :)

PR #608, opened 2017-02-15
Fix sign in refresh issues

### LINKED ISSUE #414: Sign-in not working properly on Chrome
I encounter this reproducible problem on Chrome, but not Firefox or Safari.
1. Go to https://gae-init.appspot.com/
2. Log out (if logged in), and click on "gae-init" in the nav bar.
3. Click on Log in.
4. Select Facebook, authenticate with facebook if necessary

Result: Page still says Sign In. But if you refresh, you can then see that you are logged in.

lipis:
Not the prettiest, but it does the trick.

Fixes #414

### DIFF
--- a/main/auth/auth.py
+++ b/main/auth/auth.py
@@ -414,7 +414,9 @@ def signin_user_db(user_db):
   flask.session.pop('auth-params', None)
   if flask_login.login_user(flask_user_db, remember=auth_params['remember']):
     user_db.put_async()
-    return flask.redirect(util.get_next_url(auth_params['next']))
+    next_url = util.get_next_url(auth_params['next'])
+    next_url = '%s%s%s' % (next_url, '&' if '?' in next_url else '?', util.uuid()[:4])
+    return flask.redirect(next_url)
   flask.flash('Sorry, but you could not sign in.', category='danger')
   return flask.redirect(flask.url_for('signin'))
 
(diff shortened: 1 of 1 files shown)


gmist:
It's dirty, but I see no other solution :(

``` python
next_url = util.get_next_url(auth_params['next'])
next_url = '%s%s%s' % (next_url, '&' if '?' in next_url else '?', time.time())
```

``` python
from uuid import uuid4 as uuid
import time
import timeit

def a():
    '%s' % uuid().hex[:4]

def b():
    '%s' % time.time()

print timeit.timeit('a()', 'from __main__ import a', number=10000)
print timeit.timeit('b()', 'from __main__ import b', number=10000)
```
results:
```
0.147270917892
0.00507283210754
```

rbubley:
There are trade-offs to be made here. If there was the simpler:
`next_url = '%s%s' % (next_url, '&loggedin' if '?' in next_url else '?loggedin)`

It would be faster, but would bump into this bug only if a user logged in and out with multiple userids. I suspect this multiple userid issue is an issue only for testers not real users.

gmist:
``` python
next_url = '%s%sloggedin=%s' % (next_url, '&' if '?' in next_url else '?', user_db.id)
```

lipis:
I wanted to keep it as short as possible and exposing the ID into the browser history.. might not be OK in some cases :) But based on that.. what about something like that:

```python
    next_url = '%s%s%s' % (next_url, '&' if '?' in next_url else '?', user_db.key.urlsafe()[-4:])
```

NoahWilcox:
It works perfectly for me now. Thank you!

lipis:
I'm going with this! :)

PR #615, opened 2017-02-22
Remove prefetch link on auth

lipis:
Related to #608


lipis:
Can anyone verify this?! Explains quite a few things.. why chrome was behaving like that :)

gmist:
it seems to solve the problem

lipis:
👍

mdxs:
Seems to work for me, but I had issues to reproduce the error; so I might just be on the happy path ;-)

lipis:
Thanks @mdxs :) If it works.. then we are good.. hehe

PR #617, opened 2017-02-22
Updated requirements

mdxs:
Updated requirements with newer versions of Flask, unidecode, and webargs


lipis:
ooo.. how did we miss the Flask update :D 😵

rbubley:
Maybe [requires.io](https://requires.io/) would help?

PR #620, opened 2017-03-06
Refactored email verified and token approach

mdxs:
Removed ambiguity about `verified` and `token` by making it explicit; using `email_verified` and `email_token` to make clear that these are about email.

This should make it easier for users of gae-init to integrate other attributes that could use verification, like SMS based tokens when verifying a mobile phone number.

### DIFF
--- a/main/auth/auth.py
+++ b/main/auth/auth.py
@@ -367,10 +367,10 @@ def form_with_recaptcha(form):
 ###############################################################################
 # User related stuff
 ###############################################################################
-def create_user_db(auth_id, name, username, email='', verified=False, **props):
+def create_user_db(auth_id, name, username, email='', email_verified=False, **props):
   email = email.lower() if email else ''
-  if verified and email:
-    user_dbs, cursors = model.User.get_dbs(email=email, verified=True, limit=2)
+  if email_verified and email:
+    user_dbs, cursors = model.User.get_dbs(email=email, email_verified=True, limit=2)
     if len(user_dbs) == 1:
       user_db = user_dbs[0]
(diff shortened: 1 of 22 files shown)


lipis:
Quite a change though :/ and most importantly it breaks the backwards compatibility.. we will have to have a task to upgrade the model if we are about to merge that one.. a bit skeptical TBH.

I'm away this week on [Next](https://cloudnext.withgoogle.com/) so I'll follow up next week.

mdxs:
Ah, I see; have been using it myself for a while... and thus had done the task to upgrade the model a while ago. I will think about how to resolve this. So consider this PR to be work-in-progress until further notice.

lipis:
A task could look something like this (in `task.py`) that is meant to run once after the upgrade..

```python
def user_upgrade(user_cursor=None):
  user_dbs, user_cursor, user_more = (
    model.User.query()
    .order(model.User.created)
    .fetch_page(config.DEFAULT_DB_LIMIT, start_cursor=user_cursor)
  )

  for user_db in user_dbs:
    if user_db.verified:
      user_db.email_verified = user_db.verified
    if user_db.token:
      user_db.email_token = user_db.token

  if user_dbs:
    ndb.put_multi(user_dbs)

  if user_cursor:
    deferred.defer(user_upgrade, user_cursor)
```

The thing is that it is quite a change.. :(

lipis:
As it will break quite a few things if someone is actually using it already.. I think we will have to pass on this one.

lipis:
I will close this one.. I hope it's OK.

PR #625, opened 2017-03-16
Solarized CSS Colors

lipis:
http://ethanschoonover.com/solarized


lipis:
Not the most useful thing these days..! and this is another alternative: https://yeun.github.io/open-color/

Issue #420, opened 2015-10-12
(Help Wanted) PyCharm Unresolved References


CrAzE124:
Hi there, 

I've recently started playing around with Python webdev (coming from a PHP background), and started building a project with gae-init. Working really well, I must say, but it's really irritating to not have autocomplete and what not for some imported modules. 

Let's take for example the `user.py` file in `main.api.v1` - It works fine, but it PyCharm gives me unresolved references to most of the imports - all of them in fact, except `ndb` and `absolute_import`. 

I've attached an image for reference.

Any idea what to do here? As I said, just started playing around with Python!

Thanks for any help!
![pycharm](https://cloud.githubusercontent.com/assets/6757966/10438946/05afc262-7137-11e5-8832-a2d3ecff8018.png)

gmist:
Just mark the `main` directory as `source folder` in the project settings.
<img width="707" alt="screen shot 2015-10-13 at 00 23 21" src="https://cloud.githubusercontent.com/assets/47926/10439203/0b26c8d0-7141-11e5-9f3a-d4b3a41951e5.png">
<img width="250" alt="screen shot 2015-10-13 at 00 22 17" src="https://cloud.githubusercontent.com/assets/47926/10439217/19713358-7141-11e5-958d-96762c27af3b.png">

CrAzE124:
@gmist Thanks, that solved most of the issues, although (in this example), after restart + invalidate, PyCharm still can't pick up the `restful` from `from flask.ext import restful` - any idea? I see that the `__init__.py` in `flask.ext` is supposed to import the whole `flask_restful` library, but I still get "Unresolved Reference" for that.

lipis:
Yes.. the `from flask.ext import restful` might be a bit tricky one.. Not a PyCharm expert :) If everything fails.. we might have to file an issue to JetBrains :)

CrAzE124:
@lipis interesting to note, found [this](https://youtrack.jetbrains.com/issue/PY-7142#tab=Linked%20Issues) on the PyCharm Youtrack page. Weird that it's supported, but it doesn't work out of the box I assume? Will play around a bit and see if I can find any solution (in case this bugs the hell out of anyone else!). 

Thanks for the help (and awesome skeleton app!)

lipis:
It says that is fixed.. strange :) If you find something.. please follow back :) Thanks

mroswell:
This did the trick for me, for a lot of the unresolved references. In particular, the  
[...]

PR #626, opened 2017-03-16
Remove confusing main/__init__.py

rbubley:
Related to Issue #420, and some conversations on slack, this PR removes `main/__init__.py`, and instead allows `run.py` to more explictly load only what's needed.

Positive side effects:
* it makes it clearer to see configuration variables shared between `run.py` and the main code
* it will enable a future tidy-up of `main/config.py`
* it will make mean there is no longer an ability for PyCharm CE or developers to be confused between `main` (the module) and `main.py`


mdxs:
Shouldn't the new `main/run_helper.py` at least have some `# coding: utf-8` on the top?
And perhaps add a doc string on what the possible purpose of it is.

mdxs:
Sorry, I only now realise that it was a move... I thought the file was empty, but instead the new `main/run_helper.py` is the old `main/__init__.py`; so please ignore my comment.

lipis:
Actually just delete it and add the `__version__ = '5.8.0'` inside the `run.py` and all the problems are solved. As for the description.. we'll think something else :)

rbubley:
I can do -- but the approach I proposed is:

1.  trivially compatible with gae-init-babel, which needs to import other configuration information
2. means that the main code could have access to the version number. Not sure if this is actually useful or not, but it feels like it might be.

lipis:
I will use that there instead..

```
config = imp.load_source('config', 'main/config.py')
```

lipis:
Re. 2. Not really.. as this version has to do with gae-init it self and I believe it's irrelevant with the actual app. User's might want to keep their own versions elsewhere..

lipis:
Awesome! Thanks..

PR #648, opened 2017-04-16
Fixed caps of Trusted Hosts

mdxs:
Fixed capitalisation of Trusted Hosts on configuration page; making it more in sync with how other elements are capitalised (like Brand Name. Feedback Email, and Let's Encrypt Challenge).

### DIFF
--- a/main/model/config.py
+++ b/main/model/config.py
@@ -26,7 +26,7 @@ class Config(model.Base, model.ConfigAuth):
   recaptcha_private_key = ndb.StringProperty(default='', verbose_name='Private Key')
   recaptcha_public_key = ndb.StringProperty(default='', verbose_name='Public Key')
   salt = ndb.StringProperty(default=util.uuid())
-  trusted_hosts = ndb.StringProperty(repeated=True, verbose_name='Trusted hosts')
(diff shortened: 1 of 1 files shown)


main/model/config.py
```diff
@@ -26,7 +26,7 @@ class Config(model.Base, model.ConfigAuth):
   recaptcha_private_key = ndb.StringProperty(default='', verbose_name='Private Key')
   recaptcha_public_key = ndb.StringProperty(default='', verbose_name='Public Key')
   salt = ndb.StringProperty(default=util.uuid())
-  trusted_hosts = ndb.StringProperty(repeated=True, verbose_name='Trusted hosts')
+  trusted_hosts = ndb.StringProperty(repeated=True, verbose_name='Trusted Hosts')
```
lipis: Even though lately I prefer the sentence case we'll have to keep it consistent :)

Here is an interesting post about it.. 
https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98

mdxs: Yeah, don't mind either way as long as there is consistency.

PR #657, opened 2017-05-07
Updated PyJWT and WebArgs versions

mdxs:


mdxs:
Not sure what the question is... is my editor dropping EOL characters or something?

lipis:
what*=why..  (mobile)

if you have editorconfig plugin for your editor.. it should be fine.. it looks like EOL was missing anyway from that file :)

PR #686, opened 2017-06-16
Added link to video overview of gae-init.

asmith26:
I found the  "GDG CPH - Building Web Apps Like A Pro Using Google App Engine" talk really useful, so thought it may be worth mentioning here.

Please feel free to re-word/re-position it as necessary.

Thanks for gae-init!


lipis:
I think it's quite long and old.. and it was my first public talk about it :| I always wanted to do some smaller ones just to show the basics.. but but but.. stupid me :) I'll think about the wording and maybe will include it for now..

PR #687, opened 2017-06-17
Remove disabling of user scaling.

asmith26:
Following [this article from Google on Responsive Web Design](https://developers.google.com/web/fundamentals/design-and-ui/responsive/), it states in the TL;DR section:

>Ensure your page is accessible by not disabling user scaling.

I believe removing the `user-scalable=no` will achieve this(?)

### DIFF
--- a/main/templates/bit/meta.html
+++ b/main/templates/bit/meta.html
@@ -1,4 +1,4 @@
 <meta charset="utf-8">
 <meta name="apple-mobile-web-app-capable" content="yes">
 <meta name="apple-mobile-web-app-status-bar-style" content="black">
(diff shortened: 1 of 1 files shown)


lipis:
But using Bootstrap you need that. Try it from mobile.. with this change you will be able to zoom in, something that you don't want to do I would guess.

rbubley:
Not an expert on this, but accessibility issues are important. 

Some comments on stack exchange... 

https://ux.stackexchange.com/questions/80487/revisiting-user-scalable-no

asmith26:
I'm happy either way, just thought I'd make you aware of what I read.

lipis:
I'll close this one.. as it will ruin the mobile experience.

Issue #699, opened 2017-07-19
Use "Alternative Tokens" with flask-login


asmith26:
As explained [in the flask-login documentation](https://flask-login.readthedocs.io/en/latest/#alternative-tokens):

>Using the user ID as the value of the remember token means you must change the user’s ID to invalidate their login sessions. One way to improve this is to use an alternative session token instead of the user's ID.

In particular (as far as I can tell), Google App Engine doesn't allow you to change a user's Entity Key, which is used by gae-init for flask-login session login/management.

There's a nice example [here](https://prettyprinted.com/blog/577479/using-alternative-tokens-in-flask-login) and [here](http://thecircuitnerd.com/flask-login-tokens/) for a possible implementation, however I don't think gae-init should create a session token based on a user's password (e.g. as some apps may only using OAuth login). (I'm happy to have a go at this in the next few days :-) )

PR #701, opened 2017-07-22
Add "Alternative Token" handling for flask-login (see #699)

asmith26:


main/auth/auth.py
```diff
@@ -57,7 +57,7 @@ def key(self):
     return self.user_db.key.urlsafe()
 
   def get_id(self):
-    return self.user_db.key.urlsafe()
+    return unicode(self.user_db.session_token)
```
gmist: It seems to me that ```unicode``` call is unnecessary.

``` python
def get_id(self):
  return self.user_db.session_token  # session_token == UUID
```

asmith26: Possibly, I was just filling the [docs](https://flask-login.readthedocs.io/en/latest/#your-user-class):
>get_id()
This method must return a unicode that uniquely identifies this user, and can be used to load the user from the user_loader callback. Note that this must be a unicode - if the ID is natively an int or some other type, you will need to convert it to unicode.

gmist: +1 I really missed it

lipis:
BTW.. is this breaking the backwards compatibility?

asmith26:
Good question, yes it does - sorry I forgot to test that (I'm currently still developing a new app, but will be sure to check this for any future pull requests).

One backwards compatibility problem I see is that existing users won't have been created a (unique) `session_token` (in fact, all existing users will have the same "empty" session token which is definitely not what we won't) - you can find a possible fix in https://github.com/asmith26/gae-init/commit/7d3fa4a98c3f04269ec2e38e6e16c3b6bbc1fbaf

Nonetheless to help keep the code simple, it may be better to hold off this pull request until the next major (e.g. 6.x.x) release? Alternatively, we could label this pull request with [example](https://github.com/gae-init/gae-init/pulls?q=is%3Apr+is%3Aopen+label%3Aexample)?

lipis:
I already merged it though.. maybe we can just include the possible fix (if it actually fixes and not breaking anything) otherwise I will revert this PR.

PR #725, opened 2017-10-09
Drop CoffeeScript

lipis:


lipis:
I'm about to merge that into master.. WDYT?

benoitpointet:
One less language to master in order to use gae-init. Can't be a bad thing.

lipis:
In general even now, you don't obliged to use CoffeeScript.. you can just start writing pure JavaScript and it will work out of the box..

PR #783, opened 2018-01-05
Updated momentjs

mdxs:


lipis:
I wonder why the dependabot didn't catch that one... you should update the package-lock too

mdxs:
Okay, will need to look into that...

mdxs:
Is it perhaps because "momentjs" is in the ``bower.json`` and not in ``package.json`` (and also not in ``package-lock.json``?

lipis:
dah!! :) I was so used to that.. :)

PR #791, opened 2018-01-08
Workaround for bug in babel/uglify usage in script.coffee

mdxs:
In adding support for Babel (in #770), a subtle bug was introduced in the `gulp build` target. It caused the production (minified) version of `script.js` to receive a leading `"use strict";` clause. Which in turn produced errors on the browser console like `ext.js?20180108t144620.406792256148446707:1 jQuery.Deferred exception: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them TypeError: 'caller'` and preventing users to click on list items (on Admin based List views of models for example, such as the User List); normally used for selecting their details view.

This workaround is a quick fix, causing the `script.js` to be un-minimized / un-obfuscated.

### DIFF
--- a/gulp/tasks/script.coffee
+++ b/gulp/tasks/script.coffee
@@ -14,8 +14,8 @@ gulp.task 'script', false, ->
   .pipe $.plumber errorHandler: util.onError
   .pipe $.if is_coffee, $.coffee()
   .pipe $.concat 'script.js'
-  .pipe $.babel presets: ['es2015']
-  .pipe $.uglify()
+  #.pipe $.babel presets: ['es2015']
+  #.pipe $.uglify()
   .pipe $.size {title: 'Minified scripts'}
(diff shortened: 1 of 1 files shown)


lipis:
I don't get the error.. which Browser/OS are you using and what versions?

mdxs:
It doesn't occur when testing locally on the development server (unless perhaps you force use the minified versions to be used); but it does happen when deploying the app on Google App Engine (GAE).

I've confirmed the problem with (all 64-bit) when using it from GAE:

- Windows 10 with Chrome 63.0.3239.132
- Windows 10 with Firefox 57.0.4
- Xubuntu 16.04.3 with Firefox 57.0.4

lipis:
should be OK now..

mdxs:
Confirm fixed by #818

PR #807, opened 2018-02-01
Add UglifyES

lipis:


lipis:
Still not fixes the problem though

PR #818, opened 2018-02-18
Babel presets fix

lipis:
Should also fix the #791


main/static/src/script/common/util.coffee
```diff
@@ -43,7 +43,7 @@ window.init_time = ->
         else
           $(this).text date.fromNow()
         $(this).attr 'title', date.local().format 'dddd, MMMM Do YYYY, HH:mm:ss Z'
-      setTimeout arguments.callee, 1000 * 45
+      setTimeout temp_func, 1000 * 45
```
lipis: other ideas are welcome: https://stackoverflow.com/a/12050920

main/templates/bit/script.html
```diff
@@ -1,6 +1,6 @@
 # if config.DEVELOPMENT
-  <script src="/p/dev/script/ext.js?{{config.CURRENT_VERSION_ID}}"></script>
-  <script src="/p/dev/script/script.js?{{config.CURRENT_VERSION_ID}}"></script>
+  <script src="/p/dev/script/ext.js?{{random()}}"></script>
+  <script src="/p/dev/script/script.js?{{random()}}"></script>
```
lipis: I think app engine changed something so it wouldn't update the local version ID on every refresh..

mdxs: hmm, that could have some side effects... I would rather like to have this predictable

lipis: for localhost..? does it really matter?

mdxs: Cannot (yet) reproduce that the ``{{config.CURRENT_VERSION_ID}}`` approach doesn't work.

lipis: try to change only js/css and check if the change is actually happening with autorefresh

mdxs: You mean on localhost with the development server?

package.json
```diff
@@ -2,12 +2,12 @@
   "author": "Panayiotis Lipiridis <lipiridis@gmail.com>",
   "devDependencies": {
     "babel-core": "6.26.0",
```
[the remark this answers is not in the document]
mdxs: Trying...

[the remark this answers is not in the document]
mdxs: Received an "Error: Cannot find module 'babel-core'" on ``gulp build``, while using:
````
  "devDependencies": {
    "@babel/core": "7.0.0-beta.40",
    "@babel/preset-env": "7.0.0-beta.40",
```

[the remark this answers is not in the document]
mdxs: Perhaps due to ``eslint`` requirement... ``"babel-code-frame": "6.26.0",``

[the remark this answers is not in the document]
lipis: right.. I also tried it yesterday.. but no luck 

https://github.com/babel/gulp-babel#usage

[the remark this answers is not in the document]
mdxs: Could https://github.com/babel/babel-eslint#setup perhaps show a hint... to not use ``strict``?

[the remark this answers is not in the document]
mdxs: Just keep it ``"babel-core": "6.26.0",``

package.json
```diff
@@ -2,12 +2,12 @@
...
-    "babel-preset-es2015": "6.24.1",
+    "@babel/preset-env": "7.0.0-beta.40",
```
mdxs: Just keep it ``"babel-preset-es2015": "6.24.1",``

main/static/src/script/common/util.coffee
mdxs: I'm no Javascript expert, but could the "temp_func" in this case read something like "temp_window_init_time" or "window_init_time_func" to make it more readable?

lipis: That actually doesn't work :/

lipis: should be OK now!

PR #829, opened 2018-03-05
Fix markdown code style

gmist:
- Marker style should be `-`  (remark-lint:unordered-list-marker-style)


lipis:
I'm using [Prettier](https://prettier.io) for that with [these settings](https://github.com/gae-init/gae-init/blob/master/.prettierrc.yaml#L4).. but we could remove the `$`.

Run `npm run fix` locally.

lipis:
I'll close that one because of Prettier

PR #879, opened 2018-06-15
Change travis set-up to pull in Google SDK 

rbubley:
Now pulls in Google SDK properly and uses gcloud commands as per the documents.
Also builds using yarn rather than npm (as npm will auto-run npm test, which we don't want)


lipis:
I was also thinking of switching to yarn.. but it's quite a change in general as bunch of things in `run.py` suggesting to run npm.. Can we keep npm for now and switch to Yarn in a separate PR?

Whats the problem running `npm test`?

lipis:
#881

PR #893, opened 2018-07-01
If pip install fails, warn and do not set pip.guard

rbubley:


rbubley:
I bumped into an annoying issue when I had cloned into a path with a space in it, and didn't know what had happened until the server was running, complaining that it couldn't find Flask... Tracking it down led me to propose this improvement..

run.py
```diff
@@ -168,7 +168,10 @@ def install_py_libs():
   if not check_if_pip_should_run() and os.path.exists(DIR_LIB):
     return
 
-  exec_pip_commands('pip install -q -r %s' % FILE_REQUIREMENTS)
+  make_guard_flag = True
+  if exec_pip_commands('pip install -q -r %s' % FILE_REQUIREMENTS):
+    print('ERROR running pip install')
+    make_guard_flag=False
```
lipis: spaces

rbubley: Whoops, missed that. Fixed.

lipis:
Thank you!

PR #900, opened 2018-07-09
Add dryrun option to gulp deploy

rbubley:
This should support automated testing, and aid in debugging of deployment issues.


lipis:
Good idea!

PR #970, opened 2018-10-21
Use yarn instead of npm when installing gulp-cli

rbubley:


.travis.yml
```diff
@@ -42,7 +42,7 @@ before_install:
   - python -V
 
 before_script:
-  - npm install --global gulp-cli
+  - yarn global add gulp-cli
```
lipis: not sure if we actually need global in travis..

lipis: no I guess we need it.

PR #992, opened 2018-11-02
Update requirements in Pageres section of Test page

### LINKED ISSUE #983: Update requirements in Pageres section of Test page
Should this be updated to use yarn nowadays?
https://github.com/gae-init/gae-init/blob/c9b7d6442c95acf208e26fbb829d6b29a9b9e8a9/main/templates/admin/test/test_pageres.html#L4

and would that possibly be `yarn global add pageres-cli`?

mdxs:
Update requirements in Pageres section of Test page to show usage of {{yarn}} instead of {{npm}} nowadays. Fixes #983


mdxs:
Hmmm, more commits here than I actually wanted.

PR #1001, opened 2018-11-09
Got it to where it can deploy but cannot run successfully

supern8ent:
Basically I fixed one thing at a time. It can be deployed now, both locally and to gae, but when it tries to spin up, it chokes on `import google.appengine` raising `ModuleNotFoundError: No module named 'ez_setup'`. My notes are below... 

changed runtime to python37

`ERROR: (gcloud.app.deploy) An error occurred while parsing file: [/Users/nathan/gae/gae-init/main/app.yaml]`
libraries entries are only supported by the "python27" runtime
  in "/Users/nathan/gae/gae-init/main/app.yaml", line 72, column 1

removed "libraries" section

`ERROR: (gcloud.app.deploy) The [threadsafe] field is specified in file [/Users/nathan/gae/gae-init/main/app.yaml]. This field is not supported with runtime [python37] and can safely be removed.`

removed "threadsafe" line

`ERROR: (gcloud.app.deploy) skip_files cannot be used with the [python37] runtime. Ignore patterns are instead expressed in a .gcloudignore file. For information on the format and syntax of .gcloudignore files, see https://cloud.google.com/sdk/gcloud/reference/topic/gcloudignore.`

Removed the following from app.yaml:
```skip_files:
- ^(.*/)?#.*#
- ^(.*/)?.*/RCS/.*
- ^(.*/)?.*\.bak$
- ^(.*/)?.*\.py[co]
- ^(.*/)?.*~
- ^(.*/)?Icon\r
- ^(.*/)?\..*
- ^(.*/)?app\.yaml
- ^(.*/)?app\.yml
- ^(.*/)?index\.yaml
- ^(.*/)?index\.yml
- ^lib/.*
- ^static/dev/.*
- ^static/ext/.*\.coffee
- ^static/ext/.*\.css
- ^static/ext/.*\.js
- ^static/ext/.*\.less
- ^static/ext/.*\.json
- ^static/src/.*
```
Added main/.gcloudignore with the following:
```#*#
*/RCS/*
*.bak
*.pyc
*.pyo
*~
Icon\r
.*
app.yaml
app.yml
index.yaml
index.yml
lib/*
static/dev/*
static/ext/.*.coffee
static/ext/.*.css
static/ext/.*.js
static/ext/.*.less
static/ext/.*.json
static/src/.*
```

`ERROR: (gcloud.app.deploy) INVALID_ARGUMENT: The following fields are not allowed in app.yaml: api_version.`

removed "api_version" line from app.yaml

`ERROR: (gcloud.app.deploy) INVALID_ARGUMENT: script field for handler '/_ah/stats.*' must be set to 'auto' for runtime python37.`

removed "- appstats: on" from app.yaml

`ERROR: (gcloud.app.deploy) INVALID_ARGUMENT: script field for handler '/_ah/queue/deferred' must be set to 'auto' for runtime python37.`

removed "- deferred: on" from app.yaml

`ERROR: (gcloud.app.deploy) INVALID_ARGUMENT: script field for handler '/_ah/remote_api(/.*)?' must be set to 'auto' for runtime python37.`

removed "- remote_api: on" from app.yaml, as well as "builtins:" since there was nothing left under it

`ERROR: (gcloud.app.deploy) INVALID_ARGUMENT: script field for handler '/.*' must be set to 'auto' for runtime python37.`

removed all handlers (under "error_handler:" and "handlers:")

***And it deployed!***
But it doesn't work :-(

from logs:
`import flask ModuleNotFoundError: No module named 'flask'`

Copied requriements.txt into main/

`ModuleNotFoundError: No module named 'urlparse'`

Added urlparse to requirements.txt

Ack! pip install requriements takes so long and I don't think it's necessary anymore.
Emptied the contents of requirements.txt, but kept the contents of main/requirements.txt

`Error: Could not find a version that satisfies the requirement urlparse (from -r /var/folders/3z/_26pxzns5zg947hszws95j8h0000gq/T/tmpjOPFKK (line 11)) (from versions: )`

Oh, I see: urlparse in py2 changed to urllib.parse in py3
Removed urlparse from requirements.txt

```Error:
  File "/Users/nathan/gae/gae-init/main/util.py", line 14, in <module>
    from google.appengine.datastore.datastore_query import Cursor

ModuleNotFoundError: No module named 'google'
```

added "google-cloud==0.34.0" and "google-cloud-vision==0.34.0" to requirements.txt

`ModuleNotFoundError: No module named 'google.appengine'`

added "google-cloud-datastore==1.7.1" to requirements.txt

same error

added "google-appengine==1.5.1" to requirements.txt
had to go to https://console.cloud.google.com/cloud-build/builds/33d26dcc-7caf-4ef8-b92e-a57df654c73d?project=gae-init-py3
to see the error, `ModuleNotFoundError: No module named 'ez_setup'`

added "ez_setup==0.9" to reqs

same error

stumbled on a note on the web saying that some other package trys to use setuputils, then falls back on easy_setup
removed ez_setup, added setuputils==40.5.0

same error


main/app.yaml
```diff
@@ -1,71 +1,6 @@
 service: default
-instance_class: F1
+instance_class: F2
```
lipis: let's keep it F1

requirements.txt
```diff
@@ -1,10 +1 @@
...
-flask-login==0.4.1
-flask-oauthlib==0.9.5
-flask-restful==0.3.6
-flask-wtf==0.14.2
-flask==1.0.2
-jinja2==2.10
-pyjwt==1.6.4
-unidecode==1.0.22
-webargs==4.1.2
+
```
lipis: Why not deleting the whole file?

main/requirements.txt
```diff
@@ -0,0 +1,15 @@
...
+flask-wtf==0.14.2
+flask==1.0.2
+jinja2==2.10
+pyjwt==1.6.4
+unidecode==1.0.22
+webargs==4.1.2
+setuptools==40.5.0
+google-cloud==0.34.0
+google-cloud-vision==0.34.0
+#google-appengine==1.5.1
```
lipis: hmmm.. do we need it here?

main/.gcloudignore
```diff
@@ -0,0 +1,20 @@
...
+index.yaml
+index.yml
+lib/*
+static/dev/*
+static/ext/.*.coffee
+static/ext/.*.css
+static/ext/.*.js
+static/ext/.*.less
+static/ext/.*.json
+static/src/.*
```
lipis: is that supported in the current version? Maybe we could make this change in the master?

requirements.txt
supern8ent: It was a shortcut. I expect the right thing to do is to remove the python reqs installation from gulp, but I've never edited gulp config before.

main/requirements.txt
supern8ent: Yes. In python 3 environment you tell it what libs to use when building a container via main/requirements.txt and it uses them.

lipis: but why it's commented out?

main/.gcloudignore
supern8ent: Perhaps, but a quick google for gae .gcloudignore didn't answer the question. It's a change you have to make for python 3 environment though.

rbubley:
Yes, quite a lot to do in the rest of the code, @lipis … !

Including...
  Moving away from using google auth -- so no more leaning on  that, e.g. for the auth decorators.
  Moving from ndb to Cloud Datastore
  Moving from deferred to Google Tasks
  Moving from gae mail to an external provider
  Using a WSGI app (like gunicorn) rather that the GAE infrastructure.

Then there are bound to be some python2 vs python3 issues, but these wlll mostly be dealt with in the first instance by the automated processors (covering changes to iterators, unicode), followed by a bit of manual cleanup to make it more idiomatic/stylish again.

supern8ent:
Thanks for your advice @rbubley! The list of changes seems doable at first glance. I'm not totally aware of all of the distinctions that you mentioned, such as ndb vs. cloud datastore. Can you please link the modules that you are referring to? 

I'm considering how best to attack this. If there is a risk of either total failure or massive work with migrating any of these modules, then I think it would make sense to start with that one.

rbubley:
Hi @supern8ent, 

It's well worth familiarising yourself with this document: <https://cloud.google.com/appengine/docs/standard/python3/python-differences>

There are some projects out there to replace some of the old bits with others - some of which are referenced in the above doc.

The issue with ndb is that it is an ORM (which under the hood is the Cloud Datastore), but there is not currently a google-supported interface to it in the Python 3.7 standard environment.  (There are some third-party ones, e.g. <https://github.com/lucemia/gcloud-python-orm>, and there are five years of discussions on it here: <https://github.com/GoogleCloudPlatform/google-cloud-datastore/issues/2>).

A first step, though, would be migrating to use the datastore emulator in the dev environment -- see <https://cloud.google.com/appengine/docs/standard/python/tools/migrate-cloud-datastore-emulator>. This has the advantage that it should work for both python2 and python3 (and so could be integrated into master in due course).

main/requirements.txt
rbubley: The point is that `google-appengine` *will not* be available in Python 3.7 standard environment, so migrating will involve changing all uses of this library to something else…

requirements.txt
supern8ent: In a subsequent commit I have now taken out pip install requirements from gulp/run.py and removed requirements.txt.

main/requirements.txt
supern8ent: Yes, we will have to change all the datastore code to use `google.cloud` instead of `google.appengine`.

run.py
```diff
@@ -282,7 +219,7 @@ def check_requirement(check_func):
   if not result:
     print_out('NOT FOUND', name)
     if help_url_id:
-      print 'Please see %s%s' % (REQUIREMENTS_URL, help_url_id)
+      print('Please see %s%s' % (REQUIREMENTS_URL, help_url_id))
```
rbubley: For changes like this that would work transparently in python27 too, it would be a good idea to just make them in master, if @lipis is amenable...?

rbubley:
@supern8ent, you might want to try [futurize](https://pypi.org/project/future/) on files that error like `run.py`.

supern8ent:
> To run the project locally, we will still need to install the python modules into the virtual environment.

I don't think that's true. `gcloud` seems to be building a container from `requirements.txt`.  I have no `main/lib` and when I run locally it still gets far enough to produce the same error as when I deploy to gae: `ModuleNotFoundError: No module named 'google.appengine'`.

Travis seems to be choking though...

rbubley:
> I don't think that's true. `gcloud` seems to be building a container from `requirements.txt`. I have no `main/lib` and when I run locally it still gets far enough to produce the same error as when I deploy to gae: `ModuleNotFoundError: No module named 'google.appengine'`.
> 
> Travis seems to be choking though...

You are conflating two things. The deployed project will automatically do a `pip install -r requirements.txt`, but for the project to be tested locally, either on your own machine or by travis, the requirements have to be loaded into the `temp/venv` environment. Gulp does two things - installs them into the `temp/venv` environment and loads them into the  project for upload. We need to keep the former and get rid of the latter for the python37 variant.

The extra line in `.travis.yml` of `pip install -r requriements` is just a speed optimisation so that the modules can be cached from one travis run to the next.

supern8ent:
> You are conflating two things. The deployed project will automatically do a pip install -r requirements.txt, but for the project to be tested locally, either on your own machine or by travis, the requirements have to be loaded into the temp/venv environment. Gulp does two things - installs them into the temp/venv environment and loads them into the project for upload. We need to keep the former and get rid of the latter for the python37 variant.

> The extra line in .travis.yml of pip install -r requriements is just a speed optimisation so that the modules can be cached from one travis run to the next.

Hmm. Any idea what it's doing here when I run locally? 
`INFO     2018-11-13 22:36:19,690 instance_factory.py:211] Running /var/folders/3z/_26pxzns5zg947hszws95j8h0000gq/T/tmp1rm6JI/bin/pip install -r /var/folders/3z/_26pxzns5zg947hszws95j8h0000gq/T/tmpiA7lEf`  

Locally I have no `temp/venv` and no `main/lib`, yet it's getting as far as line 14 of `util.py`, `from google.appengine.datastore.datastore_query import Cursor`, which means it successfully imports flask and urllib.

You seem to know a lot more about gae than I do, so I will happily defer; but I'm still confused on this point.

rbubley:
Ther’s plenty I don’t know!

But here, to be more precise...

I suspect dev_appserver does the pip install locally, and that is what you are seeing here — but:

1. The functional tests do not start dev_appserver, but still need access to the modules.
2. Google recommends that local testing uses either the flask dev service (`python3 main.py`) or a WSGI server like `gunicorn` - this would require a local environment; although note that google caveat’s its advice to say that dev_appserver can still be used during 2 to 3 migration though.

lipis:
Thank you @supern8ent!

PR #1017, opened 2018-11-18
Show module versions on test page

mdxs:
This PR adds a Versions test page, which will show versions of modules used in the project; including those in the so called working set. It is worth to note that it will show differences between local testing and when deployed online.

In order to do so, it includes the Google AppEngine (GAE) built-in `setuptools` (see `app.yaml`).

In the `install_py_libs()` method of `run.py` there is a further change to allow the `pkg_resources` (of `setuptools`) to be copied into the deployed project. This is thus kind of reverting #552, but I've found that this PR helps me fix issues seen when trying to use other third party libraries (such as bleach). Previously, to use bleach (for input scrubbing), I had to pin Jinja2 to the built-in version 2.6 and that in turn required me to pin Flask to version 0.12.4. In fact, debugging to show versions in that setup showed me that the GAE with such a pinned Flask/Jinja2 configuration was using a `setuptools` version 0.6c11 and I couldn't change it at all. Therefore, I tried different methods and found that this PR provided me the solution to change Flask/Jinja2 to the latest versions via the `requirements.txt` and being able to use the bleach third-party module without issues (as now it finds the `pkg_resources` it previously couldn't).

I recall that others also found difficulty with third-party libraries before, so I hope this can also help them resolve their issues. It probably helps to see which versions of modules are loaded in the deployed instance provided. You will need to configure which modules are shown in the `versions.py` source file.

### DIFF
--- a/main/app.yaml
+++ b/main/app.yaml
@@ -13,6 +13,9 @@ inbound_services:
   - warmup
 
 libraries:
+  - name: setuptools
+    version: latest
+
   - name: ssl
     version: latest
 
--- a/main/templates/admin/test/test_versions.html
+++ b/main/templates/admin/test/test_versions.html
@@ -1,9 +1,9 @@
 <p>
   Modules and their versions, '*' indicates those in the active working set.
-  <div class="alert alert-warning {{'hide' if versions}}">
-    No data on versions
-  </div>
 </p>
+<div class="alert alert-warning {{'hide' if versions}}">
+  No data on versions
+</div>
 <table class="table table-striped table-bordered table-hover {{'hide' if not versions}}">
   <thead>
     <tr>
--- a/main/templates/admin/test/test_versions.html
+++ b/main/templates/admin/test/test_versions.html
@@ -1,0 +1,22 @@
+<p>
+  Modules and their versions, '*' indicates those in the active working set.
+  <div class="alert alert-warning {{'hide' if versions}}">
+    No data on versions
+  </div>
+</p>
+<table class="table table-striped table-bordered table-hover {{'hide' if not versions}}">
+  <thead>
+    <tr>
+      <th>Name</th>
+      <th>Version</th>
+    </tr>
+  </thead>
+  <tbody>
+    # for version in versions
+      <tr>
+        <td>{{version[0]}}</td>
+        <td>{{version[1]}}</td>
+      </tr>
+    # endfor
+  </tbody>
+</table>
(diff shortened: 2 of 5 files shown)


main/app.yaml
```diff
@@ -13,6 +13,9 @@ inbound_services:
   - warmup
 
 libraries:
+  - name: setuptools
+    version: '36.6.0'
```
lipis: can we use latest?

main/templates/admin/test/test_versions.html
```diff
@@ -0,0 +1,22 @@
...
+      <th>Version</th>
+    </tr>
+  </thead>
+  <tbody>
+  # for version in versions
+    <tr>
+      <td>{{version[0]}}</td>
+      <td>{{version[1]}}</td>
+    </tr>
+  # endfor
```
lipis: indent the for loop inside the tbody

mdxs: Done

main/app.yaml
mdxs: Done

mdxs:
In the squashed commit, some further cleanup was done. I also removed the `setuptools` from the `requirements.txt` as the current requirements will already bring a `pkg_resources` that (in my testing) appears to be compatible on Google AppEngine.

When in doubt, one could always add `setuptools==36.6.0` to `requirements.txt`; where the specific version should match the `app.yaml` latest as specified at [Built-in Third-party Libraries](https://cloud.google.com/appengine/docs/standard/python/tools/built-in-libraries-27).

run.py
```diff
@@ -176,10 +176,8 @@ def install_py_libs():
 
   exclude_ext = ['.pth', '.pyc', '.egg-info', '.dist-info', '.so']
   exclude_prefix = ['setuptools-', 'pip-', 'Pillow-']
-  exclude = [
-    'test', 'tests', 'pip', 'setuptools', '_markerlib', 'PIL',
-    'easy_install.py', 'pkg_resources', 'pkg_resources.py'
-  ]
+  exclude = ['test', 'tests', 'pip', 'setuptools', '_markerlib', 'PIL',
+             'easy_install.py']
```
lipis: Put them in one line.. and no need to squash/rebase.. I'm squashing all the PRs anyway.

lipis: ping

mdxs: Ah, sorry... was busy with other stuff last month; will try to address it today.

main/templates/admin/test/test_versions.html
```diff
@@ -0,0 +1,22 @@
+<p>
+  Modules and their versions, '*' indicates those in the active working set.
+  <div class="alert alert-warning {{'hide' if versions}}">
+    No data on versions
+  </div>
```
lipis: [`div` cannot be inside `p`](https://stackoverflow.com/a/8696078/8418)

put it in a separate block

mdxs: Okay... thought I had tested this.

PR #1021, opened 2018-11-20
Python37 2to3

supern8ent:
Fairly minimal changes for 2to3 issues spread across many files.


supern8ent:
Travis' error is quoted below. Perhaps because I previously removed the venv build from gulp?  
```
[22:30:00]  $ python3 run.py
  File "run.py", line 321, in <module>
    run()
  File "run.py", line 307, in run
    check_for_update()
  File "run.py", line 171, in check_for_update
    with open(FILE_UPDATE, 'a'):
FileNotFoundError: [Errno 2] No such file or directory: 'temp/update.json'
```

lipis:
Is it possible to fix travis or I could just merge it?

rbubley:
It's right and proper that the 3.7 line should error on travis, as we don't yet expect it to be able to build and start, and it highlights what one of the things to fix is.

With regards to this particular error, I don't think it was the `venv` that created `temp/update.json`, so this should be fixable.

supern8ent:
Only Travis is throwing that error though. I would favor worrying about getting Travis working later and focus on datastore and auth sooner. We don't really have unit tests anyway.

magic.py
```diff
@@ -1,13 +1,13 @@
...
 
 from datetime import datetime
 import argparse
-import HTMLParser
+import html.parser
 import json
 import os
 import sys
-import urllib2
+import urllib.request, urllib.error, urllib.parse
```
lipis: can we just import `urllib`, otherwise have one import per line

supern8ent: I tried, it raises  
`AttributeError: module 'urllib' has no attribute 'request'`

supern8ent:
For the error,  
`FileNotFoundError: [Errno 2] No such file or directory: 'temp/update.json'`  
it was enough to add code to `check_for_update()` to create `temp` directory if it does not already exist.

supern8ent:
After adding the `temp` folder creation, Travis is now raising   
`ImportError: Failed to import test module: quick_test`

supern8ent:
@lipis Do I need to figure out this new `quick_test` thing or could you accept the PR?

lipis:
Thanks man! More fails will test.. so we can do it in another PR.

PR #1034, opened 2018-12-04
Python37

supern8ent:
Quick change: removed quick_test.py from Travis testing. This looks like a bug in google.appengine.ext.testbed since it won't even import.


rbubley:
Rather than remove the whole of the functional tests, wouldn’t it be better to remove the calls to testbed? Since these just load the additional GAE services which need to be worked around in this branch anyway, there shouldn’t be a problem with this.

supern8ent:
Ah! Sorry I didn't read far enough to realize testbed was not necessary for all the tests. So you're saying remove the first two functions but leave the latter two, right:

```
class AppTest(unittest.TestCase):
  def setUp(self):
    # Wrap the app with WebTest's TestApp.
    self.testbed = testbed.Testbed()
    self.testbed.activate()
    self.testbed.init_memcache_stub()
    self.testbed.init_datastore_v3_stub()
    from main import app
    self.testapp = TestApp(app)

  def tearDown(self):
    self.testbed.deactivate()

  def test_index_handler(self):
    response = self.testapp.get('/')
    self.assertEqual(response.status_int, 200)

  def test_non_existent_handler(self):
    response = self.testapp.get('/not-there', status=404)
    self.assertEqual(response.status_int, 404)
```

supern8ent:
Wait, not that I look at it more closely, aren't the first two necessary for the latter two to work?

rbubley:
I was thinking just of deleting every line with “testbed” in it. The bits remaining should be adequate for testing a flask app.

supern8ent:
Weird, Travis failed saying `webtest` not found despite `webtest` being present in `test_requirements.txt`. Adding it to `main/requirements.txt` does not help.

supern8ent:
Oh, I see you guys added `pip install` to `.travis.yml` but `python37` branch hasn't been rebased to include that yet.

supern8ent:
`No module named flask` during travis test.

supern8ent:
Travis failed with 
```
$ python main/runner.py --test-path tests ${HOME}/google-cloud-sdk/
test_index_handler (quick_test.AppTest) ... ERROR
test_non_existent_handler (quick_test.AppTest) ... ERROR
======================================================================
ERROR: test_index_handler (quick_test.AppTest)
----------------------------------------------------------------------
  File "/home/travis/build/gae-init/gae-init/tests/quick_test.py", line 12, in setUp
    from main import app
  File "/home/travis/build/gae-init/gae-init/main/main.py", line 3, in <module>
    import flask
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/flask/__init__.py", line 19, in <module>
    from jinja2 import Markup, escape
  File "/home/travis/google-cloud-sdk/platform/google_appengine/lib/jinja2-2.6/jinja2/__init__.py", line 33, in <module>
    from jinja2.environment import Environment, Template
  File "/home/travis/google-cloud-sdk/platform/google_appengine/lib/jinja2-2.6/jinja2/environment.py", line 576
    info.external_attr = 0755 << 16L
                            ^
SyntaxError: invalid token
```

supern8ent:
I'm stumped. This might be a travis-only problem. I set up a basic gae python37 flask project according to https://cloud.google.com/appengine/docs/standard/python3/building-app/ and added  
`import flask`  
and  
`from jinja2.environment import Environment, Template`  
and there was no problem.

rbubley:
You have a jinja2 version issue: the SyntaxError is because you are using Jinja2 v2.6 and you need v2.7 or later for python3 compatibility.

rbubley:
Looking at your edited Travis file, you have removed the `pip install -r requirements.txt`, which would be loading a newer jinja2 version — and indeed various other python modules we will need on the Travis instance to run the flask app. I think you should reinstate the pip install line.

supern8ent:
For some reason, despite installing jinja2==2.10 via main/requirements.txt and test-requirements.txt, travis is trying to load 2.6 from google-appenging: '/home/travis/google-cloud-sdk/platform/google_appengine/lib/jinja2-2.6'.

supern8ent:
I could not get tests to pass, and I can tell why it wants to load the old jinja2 from google-cloud-sdk. I think it makes sense to take tests out for now and put them back later. Bigger fish to fry. 

Also, debugging with committing and waiting for travis to test is a painfully slow loop.

supern8ent:
Dear @rbubley and @lipis ,
I just committed code that changes local serving to use `gunicorn` instead of `gcloud/dev_appserver.py` as well as adding back installing the virtualenv in temp dir. I meant for that to be a new PR, but when I pushed upstream it went straight to this PR, which was simply for removing tests (temporarily). Please advise on how best to proceed. Sorry.

PR #1043, opened 2018-12-20
Use the Cloud Datastore Emulator

rbubley:
https://cloud.google.com/appengine/docs/standard/python/tools/migrate-cloud-datastore-emulator

This will become mandatory in due course; good to flush out any issues now.

### DIFF
--- a/run.py
+++ b/run.py
@@ -348,6 +348,7 @@ def run_start():
   run_command = ' '.join(map(str, [
     'dev_appserver.py',
     DIR_MAIN,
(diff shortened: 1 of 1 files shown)


lipis:
and it doesn't break anything with backwards compatibility in the current status.. right?

rbubley:
It doesn’t seem to — but checking that was the purpose of this PR!

lipis:
Let's see if it's going to break anything!

lipis:
Thanks!

PR #1045, opened 2018-12-23
Rollback support_datastore_emulator

gmist:
_ERROR: (dev_appserver) To use the Google Cloud Datastore emulator, a Java 8+ JRE must be installed and on your system PATH_

I think this is an unacceptable dependency for Python environment.


lipis:
I agree..

rbubley:
Yes. 

We’ll be able to hold out for a bit, but Google has said that it will become mandatory at some point. At least we have a sense now what will need to change at that point.

PR #1064, opened 2019-01-15
remove-gplus-signin

jasminegrewal:
- Google+ apis are being deprecated and hence removed the call to **G+ people api**  as per [lepture/flask-oauthlib](https://github.com/lepture/flask-oauthlib/blob/master/example/google.py). Also, the changed the logo from G+ to just G (font-awesome)

- The sign-in button still doesn't align with the branding guidelines from google.


lipis:
I guess you tried it

jasminegrewal:
yes, I tried integrating the code for button from [here](https://codepen.io/timlayton/pen/gppWqz) but was out of time to figure out on how to integrate/abstract the icon code.

jasminegrewal:
yup....it worked after deploying!

lipis:
Is there any documentation for the response object?

main/auth/google.py
```diff
@@ -55,17 +55,14 @@ def retrieve_user_from_google(response):
...
-  if 'displayName' in response:
-    name = response['displayName']
-  elif 'name' in response:
-    names = response['name']
-    given_name = names.get('givenName', '')
-    family_name = names.get('familyName', '')
+  if 'name' in response:
+    name = response['name']
+  elif 'givenName' in response and 'familyName' in response:
+    given_name = response[givenName]
```
gmist: I would prefer something like this
``` python
given_name = response.get('givenName', '')
```

lipis: Me too!

jasminegrewal: my bad....made changes and did testing. Not sure about documentation.

gmist:
> Is there any documentation for the response object?

https://developers.google.com/actions/identity/google-sign-in-oauth#implement_your_oauth_server

main/auth/google.py
```diff
@@ -55,17 +55,14 @@ def retrieve_user_from_google(response):
...
 
-  if 'displayName' in response:
-    name = response['displayName']
-  elif 'name' in response:
-    names = response['name']
-    given_name = names.get('givenName', '')
-    family_name = names.get('familyName', '')
+  if 'name' in response:
+    name = response['name']
+  elif 'given_name' in response and 'family_name' in response:
```
mdxs: This "elif" is not needed. Note that with the @gmist suggested and accepted change, the code (if this "elif" is removed) will simply try to get the "given_name" and "family_name" (or get empty strings); after which the "join"+"strip" will try its best to derive a "name".

At the moment, the "elif" (suggested to be dropped) will prevent that unless both "given_name" and "family_name" are present... and that was not what the old code did and not what I expect we should require.

If both are not present, the "elif" will be skipped and the "else" will then create a 'google_user_'-based "name".

I think you should replace that "else" with an ```if not name:``` so that any (however achieved) empty name will be avoided by assigning a 'google_user_'-based "name".

main/auth/google.py
```diff
@@ -55,17 +55,14 @@ def retrieve_user_from_google(response):
...
+    given_name = response.get('given_name','')
```
mdxs: Please add space after "," in arguments, to read like: ```..._name', '')```

main/auth/google.py
```diff
@@ -55,17 +55,14 @@ def retrieve_user_from_google(response):
...
+    family_name = response.get('family_name','')
```
mdxs: Ditto, add space after ',' in arguments.

main/auth/google.py
```diff
@@ -55,17 +55,14 @@ def retrieve_user_from_google(response):
...
     name = ' '.join([given_name, family_name]).strip()
   else:
```
mdxs: Recommend replacing with ```if not name:``` (for rationale, see comment above).

main/auth/google.py
mdxs: In fact, the current ```elif 'given_name' in response and 'family_name' in response:``` should become the ```else:``` here; as the current ```else:``` below is suggested to be changed to ```if not name:``` already.

mdxs:
@jasminegrewal Thanks for this PR, appreciated. As Nathan mentioned in the gae-init Slack channel:
> ... notification from google saying that some gae-init based apps of ours are using Google+ APIs, which will start to fail "as early as January 28, 2019".

mdxs:
With recent changes, this looks good to me from a code point of view (note: I didn't test it).

mdxs:
When locally testing this, I do still see the following on the ```/admin/auth/``` path; in the (expanded) Google section:
> Do not forget to enable "Google+ API" in your console.

This either needs to be removed or updated; in: ```/main/templates/admin/bit/google_oauth.html```

Unfortunately, I couldn't easily find how/what and lack the time to figure it out; but the advise to enable  the "Google+ API" is obviously wrong.

mdxs:
Furthermore, there are more files in which the ```fa-google-plus``` should be replaced with ```fa-google```, namely in: ```/main/templates/macro/utils.html``` part should be changed to:
```
...
  # elif auth_id.startswith('google')
    <span class="fa fa-fw fa-google" title="Google"></span>
...
```
(to stop showing "G+" on profile and user list pages)

and the ```/main/templates/admin/test/test_social.html``` the file should probably become:
```
# import 'macro/utils.html' as utils

<div class="row">
  <div class="col-sm-6">
    {{utils.signin_button('Google', 'btn-google', 'fa-google', 'javascript:')}}
  </div>
  <div class="col-sm-6">
    {{utils.signin_button('Facebook', 'btn-facebook', 'fa-facebook', 'javascript:')}}
  </div>
  <div class="col-sm-6">
    {{utils.signin_button('Twitter', 'btn-twitter', 'fa-twitter', 'javascript:')}}
  </div>
  <div class="col-sm-6 text-center">
    {{utils.signin_button('Google', 'btn-google', 'fa-google', 'javascript:', True)}}
    {{utils.signin_button('Facebook', 'btn-facebook', 'fa-facebook', 'javascript:', True)}}
    {{utils.signin_button('Twitter', 'btn-twitter', 'fa-twitter', 'javascript:', True)}}
  </div>
</div>
```
to also remove "G+" from the test page.

lipis:
Thanks @mdxs! You have some good points there..

@jasminegrewal do you mind fixing those issues as well?

jasminegrewal:
@mdxs Thank you!

lipis:
@mdxs can you take another look? If nothing else.. I think we can merge that one.

mdxs:
Looks good to me.

lipis:
Thank you @jasminegrewal and @mdxs!

PR #1072, opened 2019-01-21
Fix login via Yahoo

mdxs:
Recently noticed that login via Yahoo was no longer working (don't know when it broke), probably because the [Yahoo Query Language (YQL)](https://developer.yahoo.com/yql/) API has been retired:

> Important EOL Notice: As of Thursday, Jan. 3, 2019, the YQL service at query.yahooapis.com will be retired. This will impact users of datatables.org as well as developers who creates features using this YQL service.

Therefore, this PR aims to restore the login functionality using the Explicit Grant Flow described in the [Yahoo OAuth 2.0 Guide](https://developer.yahoo.com/oauth2/guide/) and basing the implementation a lot on our it is done for Reddit in `gae-init`. Now, I'm no OAuth expert, so I hope it is sufficient as-is.

I'm happy to report that we can still get full name and email addresses when users sign-up via Yahoo (which does require use of the "Profiles (Social Directory)" permissions with the "Read/Write Public and Private" access option. I've tried with only the "Read Public" access option, but then we only get the nickname (which can be used as username and full name, similar like we do for Twitter).

Note that I've not tried "Read/Write Public" to see if that would be sufficient; as far as I have ever used Yahoo in the past, I recall that the (now obsolete) YQL method also required "Profiles (Social Directory)" permissions with the "Read/Write Public and Private" access option.

### DIFF
--- a/main/auth/yahoo.py
+++ b/main/auth/yahoo.py
@@ -2,48 +2,79 @@
 
 from __future__ import absolute_import
 
+import base64
 import flask
 
+from flask_oauthlib import client
+from werkzeug import urls
+
 import auth
+import config
 import model
 import util
 
 from main import app
 
+
 yahoo_config = dict(
-  access_token_url='https://api.login.yahoo.com/oauth/v2/get_token',
-  authorize_url='https://api.login.yahoo.com/oauth/v2/request_auth',
-  base_url='https://query.yahooapis.com/',
-  consumer_key=model.Config.get_master_db().yahoo_consumer_key,
-  consumer_secret=model.Config.get_master_db().yahoo_consumer_secret,
-  request_token_url='https://api.login.yahoo.com/oauth/v2/get_request_token',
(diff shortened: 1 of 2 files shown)


main/auth/yahoo.py
```diff
@@ -63,17 +94,23 @@ def retrieve_user_from_yahoo(response):
...
+
   return auth.create_user_db(
     auth_id=auth_id,
-    name=' '.join(names).strip() or response['nickname'],
-    username=response['nickname'],
+    name=name,
+    username=username,
     email=email,
-    verified=bool(email),
+    email_verified=bool(email),
```
lipis: Should remain `verified`.. no?

mdxs: Ah, yes... my mistake.

mdxs: Resolved

lipis:
Thanks man!

PR #1099, opened 2019-02-17
Tactical fix to oauth version compatibility issue (Fixes #1098)

### LINKED ISSUE #1098: Current code not compatible with oauthlib v3
Issue separated out from Issue #1097.

Solution proposed by @supern8ent:

> add these to `requirements.txt`:
```
oauthlib==2.1.0
requests-oauthlib==1.1.0
```

rbubley:


lipis:
Hmm.. I have the automatic merge.. I guess I should stop it then. WDYT?

mdxs:
Don't know if you can select for which libraries automatic merge is (not) to be done... if that is possible: I guess that should indeed be stopped for these two (perhaps only `request-oauthlib` pinning suffices) at the moment.

rbubley:
@lipis can do that over in #1103, with a dependabot invocation like `@dependabot ignore this dependency`

lipis:
done

PR #1114, opened 2019-03-11
added possibly correct explanation as comment

brubakerjeff:
Hi,

Let me know if this comment is correct. Trying to think of the best way to ask this and feel this is the most productive.

### DIFF
--- a/main/appengine_config.py
+++ b/main/appengine_config.py
@@ -5,7 +5,7 @@ import sys
 
 from path_util import sys_path_insert
 
-
+# if lib.zip is present use lib.zip, otherwise compile it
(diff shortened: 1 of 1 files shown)


mdxs:
Well, it is best to ask via Slack or via an Issue; using a PR to ask clarification is not correct.

@lipis Recommend to close this PR without merging.

To the question itself: the comment is incorrect, the lib.zip is built differently (via gulp tasks), whether to use it or not depends on whether you're running a local development server or are running on the Google servers; that is what the `if` is trying to figure out.

lipis:
Yes.. the comment is wrong.. and I hope @mdxs's answer answers your question.

lipis:
also it's not a big issue opening a PR with a question.. it's more for not spending your time in building something that might be already solved :)

PR #1117, opened 2019-03-16
added comment for lib.zip

brubakerjeff:
Based upon our conversation please let me know if this comment is acceptable.

### DIFF
--- a/main/appengine_config.py
+++ b/main/appengine_config.py
@@ -5,7 +5,8 @@ import sys
 
 from path_util import sys_path_insert
 
-
+# lib.zip has been precompiled from the gulp tasks when running inside of the Google Environment 
+# otherwise compile it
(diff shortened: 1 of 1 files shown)


lipis:
I'm just not sure if we really need a comment here..

brubakerjeff:
Coming from a non author working on derived work I can stare I have spent over two years not knowing how this worked. Wether or not a comment here is appropriate is debatable but I recommend a documentation update somewhere for the next person.

mdxs:
Rather than a PR to add this as a inline comment, perhaps PRs to improve the surrounding documentation?

As it is, the comment on the else is actually incorrect.

brubakerjeff:
agreed for updating documentation and I see where it is maintained now. https://github.com/gae-init/gae-init-docs

PR #1204, opened 2019-07-19
More fixes to travis

rbubley:


rbubley:
So the version of node v8.9.1 we have is incompatible with this version of eslint:

eslint@6.0.1: The engine "node" is incompatible with this module. Expected version "^8.10.0 || ^10.13.0 || >=11.10.1".

rbubley:
Ok, so if we have node 12 we need gulp 4.
Probably easiest to go to node 10.13.0 for now as I suspect we will
Have other compatibility issues with gulp 4...

.travis.yml
```diff
@@ -28,17 +28,18 @@ before_install:
...
-  - pip install -r test-requirements.txt
-  # eslint requires node ^6.14.0 || ^8.10.0 || >=9.10.0 -- but doesn't actually need it.
-  # trusty (as used on travis) comes with 8.9.1
-  - yarn install --ignore-engines
+  - pip install --upgrade -r test-requirements.txt
+  - yarn install
+
+before_script:
+  - yarn global add gulp-cli
+  - gulp -v
```
lipis: we can remove this now

rbubley: We display versions of all sorts of things for the benefit of the logging -- in theory helps debug travis issues. I'd rather leave the `gulp -v` in.

PR #1209, opened 2019-07-23
Travis fix (node version issue)

rbubley:


lipis:
Let's see

rbubley:
I see green ticks again!

lipis:
We need to update to gulp 4 in order to use Node 12.. and as a first step I think we should switch to JS instead of coffeescript for the tasks.. or maybe not. But either case it's quite a big task and it seems that @gmist is too busy :(

PR #1217, opened 2019-07-25
Drop CoffeeScript from Gulp tasks

lipis:
### Not working
- [ ] `gulp build`


rbubley:
The travis logs seem to show a race condition where `copy_bower_files` is running before `bower install`.

lipis:
Any idea why?

lipis:
@gmist?

gulp/tasks/dep.js
```diff
@@ -0,0 +1,67 @@
...
+    {
+      cmd: cmd,
+      match: /bower.json$/,
+    },
+  ];
+  gulp
+    .src('bower.json')
+    .pipe($.plumber())
+    .pipe($.start(start_map));
+});
```
rbubley: The gulp docs on start are missing, but I suspect re-writing this gulp task so as to not use `start` will fix the problem.

lipis: can you submit a PR?

rbubley: Sorry, not on this one: I've never properly got to grip with `gulp`.

rbubley:
Think you can close this now.

PR #1379, opened 2020-03-01
Add dependabot hint

rbubley:
In theory this should make dependabot only look for upgrades (in `requirements.txt`) that are Python 2.7 compatible.


rbubley:
(See https://github.com/dependabot/dependabot-core/blob/5c2dcd15a30d65a9b45425ba93addbfa336b6c32/python/spec/dependabot/python/update_checker/pip_version_resolver_spec.rb#L78)

lipis:
Didn't find anything in the settings of the Dependabot..

lipis:
Didn't notice that it was a PR :)

PR #1388, opened 2020-03-10
GitHub Auth optimization for retrieving user details

mdxs:
Faster return if already known, and reduce duplicate code for obtaining email when new user.


lipis:
Actually on a second look.. I like the `return user_db or auth.create_user_db(` better.. could you revert that part?

mdxs:
Sure, I could... but I thought the explicit "this is known, here you have it; otherwise..." was more consistent among all the methods. Note that in the otherwise part there typically (depending on the Auth method/provider) there are some things that are collected from the provider response; which is not needed if the user is already known.

But if you want mixed structures between different Auth methods... your call.

PR #1389, opened 2020-03-10
Twitter Auth optimization for retrieving user details

mdxs:
Faster return if already known, and removed typo empty line.

### DIFF
--- a/main/auth/twitter.py
+++ b/main/auth/twitter.py
@@ -27,7 +27,6 @@ twitter_config = dict(
 twitter = auth.create_oauth_app(twitter_config, 'twitter')
 
 
-
 @app.route('/api/auth/callback/twitter/')
 def twitter_authorized():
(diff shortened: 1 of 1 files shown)


main/auth/twitter.py
```diff
@@ -48,7 +47,9 @@ def signin_twitter():
 def retrieve_user_from_twitter(response):
   auth_id = 'twitter_%s' % response['id_str']
   user_db = model.User.get_by('auth_ids', auth_id)
-  return user_db or auth.create_user_db(
+  if user_db:
+    return user_db
+  return auth.create_user_db(
```
lipis: isn't that essentially the same thing?

mdxs: It is, and my objective is to make the Auth methods similar, so it is easier to see what is done for each; see my comment at https://github.com/gae-init/gae-init/pull/1388 for my considerations.

PR #1393, opened 2020-03-10
Fix enum

### LINKED ISSUE #1392: Trouble running gulp for local testing with grpcio in app.yaml
With the introduction of `grpcio` inside `app.yaml` in #1330 it seems that I now get constant:

> ImportError: No module named enum

When trying to run a local `gulp`.

Initially I thought my SDK was not compatible... but with some trial and error it seems to boil down to this addition that seems to cause a problem for me.

mdxs:
Originally, I thought that adding `enum34` might resolve #1392, however some trial and error showed that I could get a local `gulp` run working by removing `grpcio` from `app.yaml` (at least while running an old SDK 247.0.0 to which I went back trying stuff...).

### DIFF
--- a/main/app.yaml
+++ b/main/app.yaml
@@ -18,8 +18,6 @@ libraries:
 
   - name: ssl
     version: latest
-  - name: grpcio
-    version: 1.0.0
 
 error_handlers:
   - file: templates/error_static.html


mdxs:
Checked with a `gcloud components update` taking me to 284.0.0 and that now works as well with the removal of `grpcio` from `app.yaml`. So, I will squash the commit in order to make it easier to merge, as I think it is needed to keep things running (at least locally on/inside an Ubuntu 18.04 VM).

PR #1440, opened 2020-05-04
Make error reporting more robust

rbubley:
In some cases (where Exception objects do not have a code or description), we could create a KeyError when trying to report the error. This PR aims to fix that issue.


lipis:
that's a good one! Thanks

PR #1442, opened 2020-05-05
Don't run reload while initializing run

rbubley:
Previously, running gulp on a fresh instance would keep firing "reload". In this PR, "watch" doesn't start happening until after the build phase.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -195,6 +195,11 @@ gulp.task('browserSync', () => {
   });
 });
 
+gulp.task(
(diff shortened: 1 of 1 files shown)


lipis:
Is that fixing all the duplicate tasks that spamming the output?

rbubley:
It's fixing the gazillion "starting "reload" tasks. With those gone, it'll be easier to see if there are other duplicate tasks...

lipis:
Awesome!

PR #1450, opened 2020-05-06
Fix microsoft login

rbubley:


rbubley:
From my testing, the existing code in master just errors - this seems to fix.

main/auth/auth.py
```diff
@@ -344,9 +344,12 @@ def signin_oauth(oauth_app, scheme=None):
...
     flask.session.pop('oauth_token', None)
     save_request_params()
-    return oauth_app.authorize_redirect(flask.url_for(
+    redirect_uri = flask.url_for(
       '%s_authorized' % oauth_app.name, _external=True, _scheme=scheme
-    ))
+    )
+    if oauth_app.name == 'microsoft':
+        redirect_uri = redirect_uri.replace('127.0.0.1', 'localhost')
+    return oauth_app.authorize_redirect(redirect_uri)
```
rbubley, line 352: This change is because from the dev_appserver,  flask sends the redirect url as 127.0.0.1; but Microsoft insists on only permissioning localhost.

lipis:
Sounds good to me!

PR #1451, opened 2020-05-06
Re-adding LinkedIn OAuth2 support

mdxs:
This re-adds support for LinkedIn OAuth2 for login/signup; which was a capability lost when `gae-init` was refactored to use Authlib.

Note that this uses a compliance fix, based on information gathered from https://github.com/authlib/loginpass/blob/master/loginpass/linkedin.py

Which turned out to be more helpful than https://docs.authlib.org/en/latest/client/oauth2.html?highlight=linkedin#compliance-fix-for-non-standard that I also tried to follow/add to this.

I'm no OAuth2 wizard, so I just struggled until it worked for me... mixing and matching code snippets and trying out different combinations.

After I got it working, I also tried to cancel both at the login or at the approve access rights steps of LinkedIn and added handling for these two cancellation points as well.

### DIFF
--- a/main/auth/linkedin.py
+++ b/main/auth/linkedin.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import
 
 import flask
+import json
 
 import auth
 import config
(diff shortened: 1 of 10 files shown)


main/auth/linkedin.py
```diff
@@ -11,40 +12,48 @@
 
 from main import app
 
+
+def linkedin_compliance_fix(session):
```
lepture, line 16: there is no need for this `linkedin_compliance_fix` with Authlib 0.14.3 now.

mdxs: Not needed, or is it breaking something?

As this PR would make it work in the first place, I would like it to be merged as-is; after that we can see if there are further improvements possible. Simply removing this line (or code block) will not work... it is used somewhere else.

Alternatively, if this is causing a problem (with Authlib to keep this PR as-is): another PR using the newer Authlib that re-enables Linkedin as a login method is an option (and then this PR could be closed without merging).

lepture: @mdxs this `linkedin_compliance_fix` won't break anything.

lipis:
Can you also have a link somewhere that we can test it?

mdxs:
@lipis: I will send you a link via Slack to test this.

main/auth/linkedin.py
mdxs: Thanks, then I will leave it as-is for now; then see if we can indeed simply it later.

mdxs: Actually, just implemented suggestion of @lepture (to drop the linkedin_compliance_fix as it is not needed with Authlib 0.14.3 anymore). Squashed the commit log, to simplify merge.

mdxs:
You can use https://mdxs-dev.appspot.com/ for testing.

PR #1452, opened 2020-05-06
Handle cancel of access rights in Microsoft OAuth2 sign-up

mdxs:
Handle cancellation of allowing access rights in Microsoft OAuth2 sign-up process. While others (such as LinkedIn) may allow even to cancel the login (during the new user sign-up process), I couldn't find that option in the process flow. It is however the reason why this PR has a `in` test rather than a `==` test; to make it similar to the LinkedIn PR in #1451 - and I wouldn't be surprised if Microsoft would add it in the future.

### DIFF
--- a/main/auth/microsoft.py
+++ b/main/auth/microsoft.py
@@ -26,6 +26,10 @@ microsoft = auth.create_oauth_app(microsoft_config, 'microsoft')
 
 @app.route('/api/auth/callback/microsoft/')
 def microsoft_authorized():
+  err = flask.request.args.get('error')
+  if err in ['access_denied']:
+    flask.flash('You denied the request to sign in.')
+    return flask.redirect(util.get_next_url())
   id_token = microsoft.authorize_access_token()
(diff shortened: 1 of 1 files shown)


mdxs:
For those that want to check/test this process, here are some notes that I found hard to find...

When already signed-up and using Microsoft with `gae-init`, it is required to revoke `your gae-init app`'s third-party access to your Microsoft account:

1. Go to `account.microsoft.com/account` and log into your account.
2. Click on Privacy at the top of the page (or go to https://account.microsoft.com/privacy/)
3. Scroll down to the "Other privacy settings" section.
4. Click on "Apps and services that can access your data" in the "Apps and services" sub-section.

Perhaps the link https://account.live.com/consent/Manage may provide a more direct route, but you may need to be logged into your Microsoft account for it to work.

There, you can spot `your gae-init app` and edit/remove it.

Then go to your `gae-init` derived web application and sign-up with Microsoft.

PR #1476, opened 2020-06-08
Fixed extra spaces before def and import order

mdxs:
Removed extra spaces in front of a `def` and also changed import order to what is more commonly used in gae-init (first some full imports from common libs, followed by `from` imports from common libs, then full imports from project files, and finally `from` imports from project).


mdxs:
Hmm, squashing the commit failed... will re-post a PR and close this one.

Issue #1556, opened 2020-10-18
Import error with latest changes


mdxs:
Somehow recent changes cause an `ImportError: No module named cryptography.hazmat.bindings._padding` when trying to run gae-init locally in development mode.

mdxs:
I think this is due to `authlib` being set > 0.14.3 as this appears to show the issue; I've tried 0.15.2 and 0.15.1 and both show this issue when doing a `gulp reset`, `yarn install` and `gulp` locally.

PR #1595, opened 2020-12-24
Python3 Roadmap: Change treatment of requirements.txt / lib in line with migration guide

### LINKED ISSUE #596: Use official vendor library instead of sys.path manipulation
https://github.com/gae-init/gae-init/blob/master/main/appengine_config.py uses sys.path manipulation which doesn't always work correct, suggest using `google.appengine.ext.vendor` that handles `lib` as a site-packages directory.

rbubley:
This PR removes lib.zip, and much of the logic around creating it. 
In its place, it directly creates lib/ (which gets uploaded) and lib-dev/ (which doesn't but is used for the dev server.

This:
1. Fixes various obscure issues that arise when trying to use google-cloud-* modules, which are needed in the migration.
2. Will make ease the path to python 3, where GAE automatically loads requirements.txt for us.
3. Inter alia, fixes #596 and  #1556


rbubley:
I'm tempted to merge this now. Anyone object?

Issue #1720, opened 2021-08-11
Improve cleanup in development


mdxs:
It appears that the `main/libdev` folder is not cleaned up with a `gulp reset` as I would expect. This should probably be part of the `clean:venv` target, as included within a `gulp reset` command. The objective should be to remove that `main/libdev` folder.

mdxs:
Also found that https://github.com/gae-init/gae-init/blob/1b8884cda3d77498f181a5a21fc2b101be51cd92/main/app.yaml#L72 seems to be an attempt to exclude `main/libdev` from uploading to GAE when deploying. However, as-is that line doesn't do anything as there is no `main/lib-dev` folder (note the dash); and actually excluding the `main/libdev` there will make it impossible to run gae-init locally in development mode.

PR #1723, opened 2021-08-11
Improved cleanup in development

mdxs:
Improved cleanup in development, so that the `main/libdev` folder is removed on a `gulp reset` (within the `clean:venv` task).

Also removed references and tools for the now obsolete `lib.zip` approach (which #1595 addressed), including dropping of the `gulp-zip` in the `devDependencies` of `package.json` as this is no longer needed.

In the `app.yaml` there used to be an attempt to avoid the `main/libdev` folder from being deployed into GAE. However, due to a typo in `app.yaml` (using `lib-dev` in the RegEx) this actually never worked anyway. When I tried to remove the typo dash, that caused an issue when trying to run the local development server with `gulp`. Perhaps there is a way to have the libdev folder elsewhere in the local development environment, but for now it removes something that didn't work anyway (and thus the end result is the same: the `main/libdev` is deployed, as it already was).

### DIFF
--- a/gulp/paths.js
+++ b/gulp/paths.js
@@ -19,8 +19,8 @@
   paths.dep.py_guard = `${paths.temp.root}/pip.guard`;
 
   paths.py.lib = `${paths.main}/lib`;
-  paths.py.lib_file = `${paths.py.lib}.zip`;
-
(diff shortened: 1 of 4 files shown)


mdxs:
Fixes #1720

